Issue #4, opened 2026-05-03
[General] Basic CI Pipeline


ladu-tu:
A CI pipeline is needed for automatic builds and unit testing.

**Related user stories:**
None.

**Acceptance criteria:**
- [ ] A basic CI pipeline for automatic testing is implemented.

**Tasks**
- [ ] Add a CI pipeline .yml in `.github/workflows`

PR #13, opened 2026-05-06
feat(ci): add monorepo-aware CI pipeline + repo guardrails

DoPri:
## What does this PR do?

Sets up the first proper CI pipeline for the repo, plus the repo hygiene that goes with it. Monorepo-aware (paths-filter), matrix-ready for the upcoming client / GenAI services, and keeps actions/hooks up to date.

Pulled out CD on purpose - this PR is CI-only. Image push to GHCR happens on merge to main, but actual k8s deploy will land in a separate workflow once we have a target cluster.

Also includes follow-up CI maintenance:
- fixed Trivy action ref to a resolvable tag (`aquasecurity/trivy-action@v0.36.0`)
- bumped workflow actions in `.github/workflows/ci.yml` and `.github/workflows/codeql.yml`
- moved OpenAPI lint job to Node 24 in CI

## Workflows added

- ci.yml
  - changes - detect which paths a PR touches, gate everything else on it
  - lint-openapi - redocly lint on api/openapi.yaml (recommended ruleset, slightly relaxed where we don't have prod URLs yet)
  - lint-dockerfiles - hadolint matrix (one entry per Dockerfile)
  - lint-workflows - actionlint to catch typos in our own workflow files
  - build-spring - gradle build + test, matrix-ready, with build cache, uploads test reports + jar
  - build-image - buildx + docker/build-push, push to ghcr only on main, trivy scan with SARIF upload
  - compose-smoke - bring stack up via docker compose, curl /hello, tear down
  - ci-summary - single aggregate gate so branch protection only needs one required check
- codeql.yml - Java SAST, runs on PR + push + weekly cron

## Repo guardrails

- dependabot.yml for github-actions, gradle, docker
- .pre-commit-config.yaml + .yamllint.yaml so local hooks match CI
  - pre-commit-hooks: v6.0.0
  - yamllint: v1.38.0
  - hadolint: v2.14.0
  - actionlint: v1.7.12
  - replaced broken Redocly upstream hook with a local hook:
    `npx --yes @redocly/cli@2.30.3 lint api/openapi.yaml`
- redocly.yaml + .hadolint.yaml so the same rules run locally and in CI
- PR template, issue templates, CODEOWNERS skeleton
- .dockerignore for the spring service to keep build context lean

## Docs

- Updated README pre-commit section with the actual active hooks
- Added manual command: `pre-commit run --all-files`

## Drive-by fix

Patched api/openapi.yaml just enough to pass redocly recommended: server, license, operationId, ApiError envelope, explicit empty security on the public hello endpoint. Otherwise the very first PR running this CI would fail.

## Notes for the reviewer

- Branch protection: once this is green on main, please require ci-summary as the only check. Saves having to update protection every time we add a job.
- Trivy is set to exit-code 0 on purpose - we want vuln visibility before we start failing PRs on it. Flip that to 1 once base images are pinned and clean.
- Image push uses GITHUB_TOKEN to ghcr.io so no extra secret needed.
- I deliberately did not add CD here. Separate PR once we agree on the target cluster (Rancher vs Azure).

Ref. Issue #4

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,11 @@
+# Default owners for everything in the repo. Adjust per subsystem
+# once each team member's GitHub handle is wired up here.
+*               @AET-DevOps26/team-bnd
+
+# Subsystem ownership (uncomment and fill in once roles are final)
+# /services/spring/    @teammate-server
+# /services/genai/     @teammate-genai
+# /web-client/         @teammate-client
+# /api/                @AET-DevOps26/team-bnd
+# /.github/            @AET-DevOps26/team-bnd
+# /infra/              @AET-DevOps26/team-bnd
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,11 +1,5 @@
-# Default owners for everything in the repo. Adjust per subsystem
-# once each team member's GitHub handle is wired up here.
-*               @AET-DevOps26/team-bnd
+*					@BjarneHa @DoPri @ladu-tu
 
-# Subsystem ownership (uncomment and fill in once roles are final)
-# /services/spring/    @teammate-server
-# /services/genai/     @teammate-genai
-# /web-client/         @teammate-client
-# /api/                @AET-DevOps26/team-bnd
-# /.github/            @AET-DevOps26/team-bnd
-# /infra/              @AET-DevOps26/team-bnd
+/services/genai/	@DoPri
+/services/spring/	@ladu-tu
+/services/client/	@BjarneHa
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,326 @@
+# Main CI pipeline.
+#
+# Goals:
+# - Fail fast and fail loud when something is broken.
+# - Stay fast on PRs by only running jobs whose paths actually changed.
+# - Stay extensible: when web-client and the GenAI service land,
+#   they slot in as new jobs / matrix entries without rewriting this file.
+#
+# CD (push images, deploy to k8s) lives in a separate workflow once we
+# have a target cluster. This file is CI-only on purpose.
+
+name: ci
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+  workflow_dispatch:
+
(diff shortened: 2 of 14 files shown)


services/spring/.dockerignore
```diff
@@ -0,0 +1,18 @@
...
+.gitignore
+.gradle
+build/
+**/build/
+**/.gradle/
+out/
+.idea/
+*.iml
+.vscode/
+.env
```
ladnik, line 12: Why are .env files exclueded in the docker build context? Currently, we do not use any env files for the spring service, but we may in the future.

.github/CODEOWNERS
ladnik, line 1: Could you please already fill in the respective code owners?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,326 @@
...
+      needs.build-spring.result != 'failure' &&
+      (needs.changes.outputs.spring == 'true' || needs.changes.outputs.docker == 'true' || github.event_name == 'push' || github.event_name == 'workflow_dispatch')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: bring stack up
+        run: docker compose up -d --build --wait --wait-timeout 180
+
+      - name: hit /hello
```
ladnik, line 283: Should we use the "Hello World" endpoint for this CI check? I could also add a different one that better suits the "check if alive" action. But that could be done in future, so I wouldn't change it in this PR.

ladnik (changes requested):
LGTM except for some minor comments.

services/spring/.dockerignore
DoPri: The .env files do not go into the Docker image. They are only loaded at runtime and then injected as environment variables.
See: https://docs.docker.com/compose/how-tos/environment-variables/best-practices/

.github/workflows/ci.yml
DoPri: Yes, we should add a k8s compatible health check endpoint. I only added this check here to get the CI passing for this PR

.github/CODEOWNERS
DoPri: Done. See 4613734

ladu-tu (approved):
LGTM

PR #36, opened 2026-05-13
Update spring service Dockerfile according to issue description.

### LINKED ISSUE #14: [Infra] Harden the Spring service Dockerfile
The current Spring Dockerfile (`services/spring/Dockerfile`) works but has a bunch of rough edges that are worth cleaning up before we add more services and have to repeat the same patterns three times. Listing them so we don't lose track.

**Related user stories:**
None directly - this is infra hardening.

**Current state:**

```
FROM sapmachine:21 AS build
WORKDIR /app
COPY app/ .
RUN ./gradlew bootJar --no-daemon

FROM sapmachine:21-jre-headless
WORKDIR /app
COPY --from=build /app/build/libs/*.jar app.jar
EXPOSE 8080
ENTRYPOINT ["java", "-jar", "app.jar"]
```

**What's wrong:**

- Container runs as root. We don't want that - our k8s manifests will need a non-root user anyway.
- Base images aren't pinned to a digest. `sapmachine:21` floats - reproducibility goes out the window.
- No HEALTHCHECK. `docker compose up --wait` then becomes meaningless and our smoke test has to retry-loop for everything.
- No OCI labels (`org.opencontainers.image.source`, `revision`, `created`, `licenses`, ...). GHCR shows them in the package UI, and Renovate / Trivy use them.
- Build stage doesn't separate dep download from build. Any code change re-downloads the whole gradle dep tree, which kills CI cache reuse.
- No `JAVA_TOOL_OPTIONS` for memory / `-XX:MaxRAMPercentage`. Containers misjudge heap on CFS-throttled k8s nodes.
- Final image still ships the full JRE. Distroless / Wolfi / `eclipse-temurin:21-jre-alpine` would shave ~150MB.
- We picked sapmachine arbitrarily - eclipse-temurin is the more common pick in production and has tighter security advisories.

**Acceptance criteria:**

- [ ] Final image runs as a non-root user
- [ ] Both base image tags are pinned by digest (`@sha256:...`) and renovate / dependabot updates them
- [ ] HEALTHCHECK present and works against actuator `/actuator/health`
- [ ] OCI labels present (source, revision, created, licenses, title, description)
- [ ] Build is split so gradle deps are cached separately from source code
- [ ] `JAVA_TOOL_OPTIONS` set with sane container-aware defaults
- [ ] Final image size measurably smaller than current (or justified if not)
- [ ] hadolint passes at `--failure-threshold warning`
- [ ] CI compose smoke test still passes

**Tasks:**
- [ ] Complete all subissues

Once this is done we can use the same template for the future GenAI service Dockerfile.

### LINKED ISSUE #29: Decide on base image (eclipse-temurin vs sapmachine) and pin by digest

### LINKED ISSUE #30: Add `useradd`/`adduser` step + `USER` directive

### LINKED ISSUE #31: Split Dockerfile so `./gradlew dependencies` runs before source COPY

### LINKED ISSUE #32: Add HEALTHCHECK (requires Spring Actuator dependency in `build.gradle`)

### LINKED ISSUE #33: Add OCI labels driven by ARGs that CI fills via `docker/metadata-action`

### LINKED ISSUE #34: Set `JAVA_TOOL_OPTIONS` env (e.g. `-XX:MaxRAMPercentage=75 -XX:+ExitOnOutOfMemoryError`)

### LINKED ISSUE #35: Verify `docker compose up` and CI smoke test still green

ladu-tu:
## What does this PR do?

Hardens the spring service Dockerfile as requested by #14. Closes #14.
Detailled changes:
- Switched from sapmachine to eclipse-temurin (reduced image size). Closes #29.
- Pinned both base images by digest
- Added non-root user (appuser:appgroup). Closes #30.
- Split build for gradle dependency caching. Closes #31.
- Added HEALTHCHECK against /actuator/health. Closes #32.
- Added OCI labels as ARGs. Closes #33.
- Added JAVA_TOOL_OPTIONS for container memory checks. Closes #34.
- Normalized line endings in [services/spring/app/gradlew.bat](https://github.com/AET-DevOps26/team-bnd/pull/36/files/92e66df38ba8216a7656ea7d8e3322f563d7e897#diff-9002b5ff25e3baa05fef307b4e9123d4259713b4de9df2f16303e131c9eead8e)

`docker compose up` still works as expected, thus #35 is also closed.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [X] CI / infra

## Definition of Done

- [x] CI is green


services/spring/app/gradlew.bat
DoPri, line 1: This diff looks scary, but it seems to just be line ending renormalization. To my understanding, the file was committed with CRLF before. Yet, our `.gitattributes` contains the rule `*.bat text eol=crlf`, which means "store as LF in the repo, convert to CRLF on checkout on Windows". So, these changes should be fine. However, they should have been explicitly mentioned in the PR description.

services/spring/Dockerfile
DoPri, line 1: The OCI ARGs are declared before any `FROM`, which means they're not in scope inside the run stage. The LABELs will always expand to empty strings. I think, you need to re-declare them after `FROM ${JRE_IMAGE}` for them to work (https://docs.docker.com/reference/dockerfile/#understand-how-arg-and-from-interact).

Yet, looking at the CI, it doesn't actually matter right now. `docker/metadata-action` + `labels` on `build-push-action` already sets the OCI labels externally, so pushed images get correct metadata regardless:

```bash
$ docker inspect ghcr.io/aet-devops26/team-bnd/spring:sha-b3a028f | grep org.opencontainers
                "org.opencontainers.image.created": "2026-05-14T13:44:31.631Z",
                "org.opencontainers.image.description": "Repository for team BND",
                "org.opencontainers.image.licenses": "",
                "org.opencontainers.image.revision": "b3a028f9064889032a1c8ce5cc72d8165df4cdcf",
                "org.opencontainers.image.source": "https://github.com/AET-DevOps26/team-bnd",
                "org.opencontainers.image.title": "team-bnd",
                "org.opencontainers.image.url": "https://github.com/AET-DevOps26/team-bnd",
                "org.opencontainers.image.version": "main"
```

So, the Dockerfile's ARG/LABEL block is effectively dead code.
I suggest to just drop them and let metadata-action handle everything without unnecessarily bloating the Dockerfile.

DoPri (changes requested):
Overall solid. Dockerfile hardening is well done (temurin, digest pinning, non-root, layer caching, healthcheck). OCI label handling should be revised and the `gradlew.bat` line ending renormalization be documented.

services/spring/app/gradlew.bat
ladu-tu: I updated the PR description.

services/spring/Dockerfile
ladu-tu: Thanks, I must have missed this!

services/spring/Dockerfile
```diff
@@ -1,12 +1,50 @@
...
+
+COPY app/src/ src/
 RUN ./gradlew bootJar --no-daemon
 
-# run
-FROM sapmachine:21-jre-headless
+# Run stage
+FROM ${JRE_IMAGE}
+
+LABEL org.opencontainers.image.source="${OCI_SOURCE}" \
```
DoPri: These `LABEL`s should also be removed.

DoPri (changes requested):

DoPri (approved):
Thanks. LGTM.

PR #68, opened 2026-05-18
Feat: Add basic user management and user API

### LINKED ISSUE #5: [Server] Basic User Management
This feature provides basic user management on the server side. This includes registration, deregistration and login. The authentication should work via JWT.

Based on the analysis object model in docs/diagrams/.

**Related user stories:**
- As a user, I want to register an account so I can use Alexandria
- As a user, I want to log in so I can access my documents
- As a user, I want to delete my account if I no longer need it

**Acceptance criteria:**

- [ ] User entity persists to database with fields from analysis object model (id, username, email, passwordHash, createdAt, preferences)
- [ ] New users can be registered through the service layer
- [ ] Passwords are hashed (bcrypt or similar)
- [ ] Users can authenticate and receive a JWT
- [ ] Users can delete their account
- [ ] Service layer has unit test coverage
- [ ] Integration tests verify database and auth operations

**Tasks:**

- [ ] Create User entity and repository
- [ ] Implement UserService with registration, auth, deletion
- [ ] Implement JWT token generation and validation
- [ ] Add unit and integration tests

### LINKED ISSUE #8: [API] Set up Spring OpenAPI generation and User Management endpoints
First part of defining the API. The springdoc-openapi should generate the OpenAPI spec from code at build time, for which Spring Boot controllers need to be annotated. Also basic user management endpoints should be implemented. 

**Acceptance criteria:**
- [ ] `./gradlew build` produces `api/openapi.yaml`
- [ ] Swagger UI accessible at `/swagger-ui.html` in dev
- [ ] User CRUD endpoints work and are documented in the generated spec
- [ ] Spec passes redocly lint

**Tasks:**
- [ ] Add `springdoc-openapi-starter-webmvc-ui` dependency to the user-service
- [ ] Configure springdoc to export `openapi.yaml` at build time (Gradle task)
- [ ] Create User entity and repository
- [ ] Implement user management endpoints (see sub-issues)
- [ ] Add OpenAPI annotations to all endpoints
- [ ] Verify generated spec is valid (`npx @redocly/cli lint`)

### LINKED ISSUE #43: [API] Add springdoc-openapi dependency and Gradle export task
Part of #8.

Add the springdoc dependency to user-service and configure a Gradle task that exports the OpenAPI spec to `api/openapi.yaml` during build.

**Acceptance criteria:**

- [ ] Running build produces a valid `api/openapi.yaml`
- [ ] Swagger UI is accessible at `/swagger-ui.html` when running locally

**Tasks:**

- [ ] Add `springdoc-openapi-starter-webmvc-ui` to `build.gradle`
- [ ] Add `springdoc-openapi-gradle-plugin` for spec export
- [ ] Configure plugin to output to `api/openapi.yaml`
- [ ] Test that `./gradlew generateOpenApiDocs` works

### LINKED ISSUE #49: [API] DELETE /api/v1/users/{id} - Delete user endpoint
Part of #8.

Implement endpoint to delete a user.

**Acceptance criteria:**

- [ ] `DELETE /api/v1/users/{id}` removes user and returns 204
- [ ] Non-existent ID returns 404
- [ ] Endpoint documented in OpenAPI spec

**Tasks:**

- [ ] Add `DELETE /api/v1/users/{id}` to `UserController`
- [ ] Return 204 on success
- [ ] Return 404 if user not found
- [ ] Add OpenAPI annotations

### LINKED ISSUE #57: [Server] Create User entity with JPA annotations
Part of #5.

Create the User domain entity based on the analysis object model (docs/diagrams/analysis_object_model.png).

**Acceptance criteria:**

- [ ] User entity compiles and is picked up by Spring Data
- [ ] All fields from analysis object model are present: id, username, email, passwordHash, createdAt, preferences
- [ ] Can persist and retrieve users via repository in a test

**Tasks:**

- [ ] Create `User` entity class in the user-service
- [ ] Add JPA annotations (`@Entity`, `@Id`, `@GeneratedValue`, `@Column`, etc.)
- [ ] Add appropriate types (UUID for id, Instant for createdAt, etc.)
- [ ] Decide on preferences storage (JSON column, separate table, or simple string for now)

### LINKED ISSUE #58: [Server] Create UserRepository interface
Part of #5.

Create the Spring Data JPA repository for User entity.

**Acceptance criteria:**

- [ ] UserRepository interface exists and extends JpaRepository
- [ ] Basic CRUD operations work (inherited from JpaRepository)
- [ ] Can find user by username and email
- [ ] Repository is autowirable in services

**Tasks:**

- [ ] Create `UserRepository` interface extending `JpaRepository<User, UUID>`
- [ ] Add `findByUsername(String username)` method
- [ ] Add `findByEmail(String email)` method
- [ ] Add `existsByUsername(String username)` and `existsByEmail(String email)` for registration validation

### LINKED ISSUE #59: [Server] Implement UserService with registration and account management
Part of #5.

Implement the service layer for user registration and account management.

**Acceptance criteria:**

- [ ] Users can register with username, email, password
- [ ] Passwords are hashed before storage (bcrypt)
- [ ] Duplicate username/email rejected with appropriate error
- [ ] Users can delete their account
- [ ] Service is injectable via constructor injection

**Tasks:**

- [ ] Create `UserService` class with `@Service` annotation
- [ ] Inject `UserRepository` and `PasswordEncoder` via constructor
- [ ] Implement registerUser() - validate uniqueness, hash password, save
- [ ] Implement deleteUser() - remove user by ID
- [ ] Implement getUserByUsername() and getUserById() for lookups
- [ ] Add proper error handling (throw appropriate exceptions)

### LINKED ISSUE #60: [Server] Implement JWT authentication
Part of #5.

Implement JWT-based authentication for the user service.

**Acceptance criteria:**

- [ ] Users can authenticate with username/email and password
- [ ] Successful auth returns a signed JWT
- [ ] JWT contains user ID and expiration
- [ ] JWT can be validated on subsequent requests
- [ ] Invalid credentials return appropriate error

**Tasks:**

- [ ] Add JWT library dependency (jjwt or spring-security-oauth2-jose)
- [ ] Create `JwtService` for token generation and validation
- [ ] Configure JWT secret via environment variable (not hardcoded)
- [ ] Implement login endpoint logic in UserService or AuthService
- [ ] Set reasonable token expiration (e.g., 24h)
- [ ] Add Spring Security config if needed for protected endpoints

### LINKED ISSUE #61: [Server] Add unit tests for UserService
Part of #5.

Add unit tests for the UserService layer using mocked repository.

**Acceptance criteria:**

- [ ] Unit tests cover registration, login, deletion
- [ ] Tests use mocked UserRepository (no real DB)
- [ ] Edge cases tested (duplicate user, wrong password, etc.)
- [ ] Tests run in CI

**Tasks:**

- [ ] Create `UserServiceTest` class
- [ ] Mock `UserRepository` and `PasswordEncoder` using Mockito
- [ ] Test registerUser() - happy path and duplicate user case
- [ ] Test login/authenticate - happy path, wrong password, user not found
- [ ] Test deleteUser() - happy path and not-found case

### LINKED ISSUE #62: [Server] Add integration tests for user persistence and auth
Part of #5.

Add integration tests that verify user persistence and authentication against a real database.

**Acceptance criteria:**

- [ ] Integration tests verify actual database operations
- [ ] User entity persists and retrieves correctly
- [ ] Password hashing works end-to-end
- [ ] JWT generation and validation works
- [ ] Tests run in CI

**Tasks:**

- [ ] Create `UserRepositoryIntegrationTest` class
- [ ] Use `@DataJpaTest` or `@SpringBootTest` with test database
- [ ] Test persist and retrieve User
- [ ] Test unique constraint violations (duplicate username/email)
- [ ] Create `AuthIntegrationTest` for login flow
- [ ] Test JWT token generation and validation

ladu-tu:
## What does this PR do?

This PR addresses issues #5 and #8, which are comprised of basic user management on the server side (with persistent storage in a PostgreSQL database) and the corresponding REST API endpoints. As for the API, this PR also implements auto-generation of the `openapi.yaml` from the spring boot code. Therefore, this is a rather large pull request :)
This closes #5.

Detailled summary:
Concerning #5:
- Added User entity. Closes https://github.com/AET-DevOps26/team-bnd/issues/57
- Added UserRepository interface. Closes https://github.com/AET-DevOps26/team-bnd/issues/58
- Added UserService. Closes https://github.com/AET-DevOps26/team-bnd/issues/59
- Added authentication via JWT. Closes https://github.com/AET-DevOps26/team-bnd/issues/60
- Added unit and integration tests for all added entities/services. Closes https://github.com/AET-DevOps26/team-bnd/issues/61 and closes https://github.com/AET-DevOps26/team-bnd/issues/62

Concerning #8 (not all subissues completed yet, later PR will address these):
- Added automatic generation of OpenAPI spec as gradle task. Closes https://github.com/AET-DevOps26/team-bnd/issues/43

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor / cleanup
- [X] Docs
- [x] CI / infra

## Definition of Done

- [x] CI is green
- [X] If the API changed: `api/openapi.yaml` is updated and lint passes
- [X] If a service was added or restructured: docker-compose still comes up cleanly (`docker compose up`)
- [X] Tests added or updated for the changed behaviour
- [X] Docs updated where it matters (README, ADRs, comments)


services/spring/app/src/main/java/com/alexandria/app/auth/SecurityConfig.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Bean
+    public PasswordEncoder encoder() {
+        return new BCryptPasswordEncoder();
+    }
+
+    @Bean
+    public SecurityFilterChain filters(HttpSecurity http) throws Exception {
+        http
+                .csrf(csrf -> csrf.disable())
```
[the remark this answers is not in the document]
ladu-tu: This is a false positive: Our API as it is currently designed (and as required by the "Best Practices Microservices" section on Artems) to use stateless JWT Bearer token authentication, not session cookies.

These bearer tokens are not automatically attached by a browser and CSRF attacks are thus not possible. Refer to this [answer](https://security.stackexchange.com/a/170414). Adding CSRF protection would defeat our stateless architecture.

ladu-tu:
The CI checks pass, finally!

The issue was that I had set the repository variables to values with quotation marks, apparently GitHub already passes them with "<value>". The resulting double quotations (""<value>"") lead to a failure in the postgres-db container, by which the spring service couldn't connect to the database and start up.

services/spring/app/src/main/java/com/alexandria/app/user/UserRepository.java
```diff
@@ -0,0 +1,18 @@
...
+@Repository
+public interface UserRepository extends JpaRepository<User, UUID> {
+    Optional<User> findByUsername(String username);
+
+    Optional<User> findByEmail(String email);
+
+    boolean existsByUsername(String username);
+
+    boolean existsByEmail(String email);
+}
```
BjarneHa, line 22: No newlines at end of java files.
Should these be added? Perhaps we should create a pre-commit hook for this.

api/openapi.yaml
```diff
@@ -1,120 +1,162 @@
...
-                type: string
-                example: Hello World!
-        '500':
-          description: Unexpected server error
+                $ref: "#/components/schemas/AuthResponse"
+        "400":
+          description: Invalid input or duplicate user
           content:
-            application/json:
+            '*/*':
```
DoPri: I don't think having a wildcard pattern instead of a proper MIME-Type is a good idea. Is there a reason for this?

api/openapi.yaml
```diff
@@ -1,120 +1,162 @@
...
+              $ref: "#/components/schemas/LoginRequest"
+        required: true
       responses:
-        '200':
-          description: Successful response
+        "200":
+          description: Login successful
           content:
-            text/plain:
+            '*/*':
```
DoPri: Same as above

api/openapi.yaml
```diff
@@ -1,120 +1,162 @@
...
-                type: string
-                example: Hello from Alexandria GenAI!
-        '500':
-          description: Unexpected server error
+                $ref: "#/components/schemas/AuthResponse"
+        "401":
+          description: Invalid credentials
           content:
-            application/json:
+            '*/*':
```
DoPri: Same as above

api/README.md
```diff
@@ -1,4 +1,7 @@
 # API
-We use [Swagger Editor](https://editor.swagger.io/) to define our API. The respective client implementations can then be generated via tools as described in the "Best Practices Microservices" section on Artemis.
+The API is generated automatically from the annotated REST endpoints in the server code.
```
DoPri, line 2: Since the GenAI component also has a REST API, but is not part of the Spring backend, how would we add its endpoints to the `openapi.yml`? We don't want `./gradlew generateOpenApiDocs` to overwrite these endpoints if we do manual changes in the yml

services/spring/app/src/main/java/com/alexandria/app/auth/SecurityConfig.java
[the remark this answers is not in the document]
DoPri: Thank you for clarifying :)

services/spring/app/src/main/java/com/alexandria/app/exception/GlobalExceptionHandler.java
DoPri, line 1: What about the other HTTP errors? Is there a catch-all for them?

services/spring/app/src/main/resources/application-openapi.properties
```diff
@@ -0,0 +1,20 @@
...
+spring.application.name=alexandria
+
+# API docs
+springdoc.api-docs.enabled=true
+springdoc.swagger-ui.enabled=true
+springdoc.swagger-ui.path=/swagger-ui.html
+
+
+# Database (in memory)
+spring.datasource.url=jdbc:h2:mem:openapi;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE
```
DoPri, line 10: We would need to change this to Postgres later, right?
What's the difference between this file and the regular `application.properties`?

services/spring/app/build.gradle
```diff
@@ -2,6 +2,16 @@ plugins {
 	id 'java'
 	id 'org.springframework.boot' version '4.0.6'
 	id 'io.spring.dependency-management' version '1.1.7'
+	id 'org.springdoc.openapi-gradle-plugin' version '1.9.0'
+}
+
+openApi {
+	apiDocsUrl = "http://localhost:8080/v3/api-docs.yaml"
```
DoPri, line 9: Is hardcoding localhost here fine?

docker-compose.yml
```diff
@@ -16,6 +17,13 @@ services:
 
   spring:
     build: services/spring/
+    container_name: spring
+    env_file: .env.config
```
DoPri, line 21: Now we have the Env variables inside here AND in the .env file. I don't really like this duplication as it does seem unnecessary.

Also, why `.env.config` here? Will this still honor a local `.env` file?

DoPri (commented):
I really like that you already wrote unit tests that work in our CI.
Before approving, however, I got some questions/uncertainties that I'd like addressed.

api/openapi.yaml
ladu-tu: I did not explicitly specify the MIME type, so this is the default generated by springdoc-openapi. I will change it.

api/README.md
ladu-tu: I think the best way is to generate two separate files ('api/spring-openapi.yaml' and 'api/genai-openapi.yaml') and then merge them using `swagger-merge` (cf. [here](https://stackoverflow.com/a/61761432)).

I think we should also add the spec generation to the CI or a git commit prehook, so the API spec is always up to date.

What do you think about this?

services/spring/app/src/main/java/com/alexandria/app/exception/GlobalExceptionHandler.java
ladu-tu: Spring Boot handles exceptions by default with a code 500 response, but I explicitly added it to the GlobalExceptionHandler.java now so it becomes clear. 

The exception handler is also not final, there may be additions later on once i expand the service further.

services/spring/app/build.gradle
ladu-tu: Yes, this is fine imo. The apiDocsUrl is only used during local development when running `./gradlew generateOpenApiDocs`

services/spring/app/src/main/resources/application-openapi.properties
ladu-tu: No, I would not change to Postgres. The application-openapi.properties is just a properties file used solely for the generation of the OpenAPI spec; springdoc-openapi spins up the spring application and fetches the API from the `/v3/api-docs` endpoint. For that, we don't need an actual persistent database, that's why we use H2 (in-memory).

docker-compose.yml
ladu-tu: Yes, docker compose loads the `.env` file in the same directory [automatically](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#env-file), but only for ${VAR} substitution *inside* the docker-compose.yml.

My idea was that the .env file contains secrets that should not be commited to the repository, respectively defined as repository secrets. That way, the docker-compose.yml works both for local development (passing of .env variables to the containers) and in the CI (where the variables will be populated with the repository secrets, not from the .env file).

The .env.config should contain variables that can/should be committed to the repository.

I see your point about the unneccessary duplication. I could add a small workaround that just generates an .env file from repository variables for the CI on the fly. However, that would need to be duplicated when adding other stages that run docker compose and would make the CI more complex.

I would keep it as is for the time being.

services/spring/app/src/main/java/com/alexandria/app/exception/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @ExceptionHandler(InvalidCredentialsException.class)
+    public ResponseEntity<ErrorResponse> handleUnauthorized(InvalidCredentialsException e) {
+        return ResponseEntity.status(HttpStatus.UNAUTHORIZED)
+                .body(new ErrorResponse("UNAUTHORIZED", e.getMessage()));
+    }
+
+    // default handler for all other exceptions
+    @ExceptionHandler(Exception.class)
+    public ResponseEntity<ErrorResponse> handleGenericException(Exception ex) {
```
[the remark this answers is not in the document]
ladu-tu: Again a false positive. Spring requires this signature for the exception handler.

services/spring/app/src/main/java/com/alexandria/app/auth/SecurityConfig.java
```diff
@@ -0,0 +1,63 @@
...
+     * All other endpoints require a valid JWT from OIDC provider.
+     *
+     * @param http HttpSecurity builder.
+     * @return Configured SecurityFilterChain.
+     * @throws Exception If configuration fails.
+     */
+    @Bean
+    public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
+        http
+                .csrf(csrf -> csrf.disable())
```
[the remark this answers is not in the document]
ladu-tu: Again a false positive. Same reasons as stated previously.

ladu-tu:
As discussed in our meeting on thursday, I switched from a custom JWT Auth to using keycloak as an OIDC provider.

To verify that the OIDC authentication works, you can e.g. test as follows:

Spin up the containers:

```bash
docker compose up -d
```
Get a token from Keycloak:

 ```bash
TOKEN=$(curl -s -X POST "http://localhost:8180/realms/alexandria/protocol/openid-connect/token" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d "grant_type=password" \
  -d "client_id=alexandria-client" \
  -d "username=testuser" \
  -d "password=testpassword" | jq -r '.access_token')
```

Try to delete a non-existant user (no other protected API endpoints exist yet). Expected: "HTTP: 403" (can't delete another user)

```
curl -H "Authorization: Bearer $TOKEN" \
  -w "\nHTTP: %{http_code}\n" \
  -X DELETE "http://localhost:8080/api/v1/users/00000000-0000-0000-0000-000000000000"
```

Verify that the user with the access token has been automatically added to the database upon their first request:

```
docker compose exec postgres-db psql -U admin -d alexandria \
  -c "SELECT id, oidc_subject, username, email FROM users;"
```

File: docker-compose.yml (hunk 9 above)
Comment (DoPri) [reply docker-compose.yml:21]: I think a separate `docker-compose.override.yml` for the CI would be a better idea, but we can do this in another PR if you feel that's too much work for now

File: api/README.md (hunk 6 above)
Comment (DoPri) [reply api/README.md:2]: Sounds good. It should be a pre-commit hook imo so that the repo always contains **one** `openapi.yml` that contains all endpoints.

**Review by DoPri — COMMENTED** [review]
Nice! Thank you for implementing the changes as discussed in our meeting so quickly.
I think @BjarneHa is still working on a review, so maybe wait for him before merging.

File: docker-compose.yml (hunk 9 above)
Comment (ladu-tu) [reply docker-compose.yml:21]: I would do that in another PR.

File: services/spring/app/src/main/resources/application.properties [hunk 12]
```diff
@@ -3,4 +3,18 @@ spring.application.name=alexandria
...
+
+# Database
+spring.datasource.url=jdbc:postgresql://postgres-db:5432/${POSTGRES_DB}
+spring.datasource.username=${POSTGRES_USER}
+spring.datasource.password=${POSTGRES_PASSWORD}
+spring.jpa.hibernate.ddl-auto=update
+
+# OIDC provider configuration
+spring.security.oauth2.resourceserver.jwt.issuer-uri=${OIDC_ISSUER_URI}
+spring.security.oauth2.resourceserver.jwt.jwk-set-uri=${OIDC_JWK_SET_URI}
```
BjarneHa, line 16: Why is this set twice?

services/spring/app/src/main/java/com/alexandria/app/user/User.java
```diff
@@ -0,0 +1,88 @@
...
+    private UUID id;
+
+    @Column(nullable = false, unique = true)
+    private String oidcSubject;
+
+    @Column(unique = true)
+    private String username;
+
+    @Column(unique = true)
+    private String email;
```
BjarneHa, line 26: Should perhaps also not be nullable

BjarneHa (changes requested):
LGTM except for some very minor points. Good work!

services/spring/app/src/main/java/com/alexandria/app/user/UserRepository.java
ladu-tu: We could add a pre-commit hook for this, but I don't think it is that relevant for this PR.

services/spring/app/src/main/resources/application.properties
ladu-tu: Good catch!

services/spring/app/src/main/java/com/alexandria/app/user/User.java
ladu-tu: As we extract username and email from the JTW, they cannot be guaranteed to not be null because the OpenID spec does not strictly enforce that these claims are not null.

BjarneHa (approved):

PR #95, opened 2026-05-27
Add POSTGRES_HOST and POSTGRES_PORT environment variables

BjarneHa:
## What does this PR do?

This PR adds `POSTGRES_HOST` and `POSTGRES_PORT` environment variables to the spring service.
These values were hardcoded previously.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] CI / infra

## Definition of Done

- [ ] CI is green
- [ ] Docs updated where it matters (README, ADRs, comments)

## Notes for the reviewer

--


docker-compose.yml
ladu-tu: Could you adapt the port mapping to use the environment variable?

docker-compose.yml
```diff
@@ -20,6 +20,8 @@ services:
     container_name: spring
     env_file: .env.config
     environment:
+      POSTGRES_HOST: "postgres-db"
+      POSTGRES_PORT: 5432
```
ladu-tu: These should be defined in .env.config and passed as environment variables here, not as explicit values.

.github/workflows/ci.yml
```diff
@@ -32,6 +32,8 @@ env:
   JAVA_DISTRIBUTION: "temurin"
   REGISTRY: ghcr.io
   IMAGE_NAMESPACE: ${{ github.repository }}
+  POSTGRES_HOST: ${{ vars.POSTGRES_HOST }}
+  POSTGRES_PORT: ${{ vars.POSTGRES_PORT }}
```
ladu-tu, line 36: Did you already set up the repository variables? I don't see them yet.

ladu-tu (changes requested):
Just some minor changes, otherwise looks good.

docker-compose.yml
BjarneHa: I decided to leave this fixed since the port for spring is also fixed in the `docker-compose.yaml` and I don't see the use-case for changing which port these two services communicate on. But I can change this if you'd like.

docker-compose.yml
BjarneHa: I decided against this since the host is only dependent on the name of the postgres container. Therefore, configuring this could *only* break things. 

As for port, see above.

docker-compose.yml
ladu-tu: You can leave it as is if you want, but it would look cleaner to have only one environment file in which everything can be configured.

docker-compose.yml
BjarneHa: Point being, `POSTGRES_HOST` and `POSTGRES_PORT` are not meant as configuration options for the `docker-compose.yaml`, but as configurations options when running the services outside of the `docker-compose.yaml` (e.g., with Helm).

ladu-tu: Ah, makes sense. Fine by me.

.github/workflows/ci.yml
ladu-tu: I added them now.

BjarneHa: Thanks :)

ladu-tu (approved):

docker-compose.yml
BjarneHa: Since the port is also visible from the outside, I added it to the `.env.config`. Thanks.

Issue #67, opened 2026-05-18
Add an example test suite


BjarneHa:
Add an example test suite for demonstration

Issue #66, opened 2026-05-18
Document the playwright setup


BjarneHa:
Document the setup in the repository's `README.md` (or a dedicated `E2E_TESTING.md`) with instructions on running tests locally.

Issue #65, opened 2026-05-18
Add Playwright to the CI workflow (GitHub Actions)


BjarneHa:
- Ensure Playwright browsers are installed during CI (`npx playwright install`).
- Add Playwright to the CI workflow

Issue #64, opened 2026-05-18
Create a basic Playwright configuration


BjarneHa:
Create a basic Playwright test configuration (playwright.config.ts) with sensible defaults (headless mode, test directory, timeout, reporters).

PR #126, opened 2026-06-12
Feat: Add setup for E2E testing with Playwright

### LINKED ISSUE #63: [Client] Set up Playwright for the client and integrate it into the CI pipeline
# Set up Playwright for the client and integrate it into the CI pipeline

## Description
Implement end-to-end browser testing for the client application using Playwright. The tests should be runnable locally and be executed automatically as part of the CI pipeline to catch regressions in UI/UX.

## Tasks
- [x] Add Playwright as a development dependency in the client project (`npm install -D @playwright/test`).
- [x] Create a basic Playwright test configuration (`playwright.config.ts`) with sensible defaults (headless mode, test directory, timeout, reporters).
- [x] Add an example test suite (e.g., navigation to the home page and verification of a critical UI element).
- [x] Update the client `package.json` with npm scripts.
- [x] Ensure Playwright browsers are installed during CI (`npx playwright install`).
- [x] Add Playwright to the CI workflow (GitHub Actions).
- [x] Document the setup in the repository's `README.md` (or a dedicated `E2E_TESTING.md`) with instructions on running tests locally.

## Acceptance Criteria
- [x] Playwright tests run successfully on a developer machine (`npm run test:e2e`).
- [x] CI workflow executes the same tests and fails on any test failure.
- [x] Browsers install automatically in CI without manual steps.
- [x] Documentation is added for developers to run/debug the tests locally.
- [x] The CI step completes within roughly five minutes.

BjarneHa:
## What does this PR do?

Adds the required setup for E2E testing with Playwright.
Closes #63, #64, #65, #66 and #67.
A CI job is added, which starts all services with docker-compose and then runs the Playwright test suite.
For local testing, npm scripts are provided, which run using a Playwright server in a Docker container.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] CI / infra

## Definition of Done

- [x] CI is green

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -502,7 +502,7 @@ jobs:
 
       - name: install playwright browsers
         working-directory: services/client
-        run: npx playwright install --with-deps chromium
+        run: npx playwright install --with-deps
 
       - name: run playwright tests
         working-directory: services/client
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -419,6 +419,113 @@ jobs:
         run: docker compose down -v --remove-orphans
 
   # ---------------------------------------------------------------------
+  # E2E tests: bring the full stack up with Docker Compose, provision a
+  # Keycloak test user via the Admin REST API, then run Playwright tests
+  # ---------------------------------------------------------------------
+  e2e:
+    name: e2e tests (playwright)
+    needs: [changes, build-spring, build-client, build-genai]
+    if: |
+      always() &&
+      needs.build-spring.result != 'failure' &&
+      needs.build-client.result != 'failure' &&
+      needs.build-genai.result != 'failure' &&
+      (needs.changes.outputs.spring == 'true' || needs.changes.outputs.client == 'true' || needs.changes.outputs.genai == 'true' || needs.changes.outputs.docker == 'true' || github.event_name == 'push' || github.event_name == 'workflow_dispatch')
+    runs-on: ubuntu-latest
+    env:
+      TEST_USER: ${{ vars.TEST_USER }}
+      TEST_PASSWORD: ${{ secrets.TEST_PASSWORD }}
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: bring stack up
+        run: docker compose up -d --build --wait --wait-timeout 180
+
+      - name: wait for keycloak realm
+        run: |
+          for i in $(seq 1 20); do
+            if curl -fsS "http://localhost/auth/realms/alexandria" > /dev/null 2>&1; then
+              echo "keycloak realm ready"; break
+            fi
+            echo "waiting for keycloak... ($i/20)"
+            sleep 5
+          done
+
+      - name: provision test user via keycloak admin api
+        run: |
+          # Obtain an admin access token from the master realm
+          ADMIN_TOKEN=$(curl -fsS \
+            -d "client_id=admin-cli" \
+            -d "username=${KC_BOOTSTRAP_ADMIN_USERNAME}" \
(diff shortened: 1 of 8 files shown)


DoPri (changes requested):
Please fix the CI failures and then re-request a review

services/client/package.json
```diff
@@ -5,13 +5,17 @@
   "scripts": {
     "dev": "vite",
     "build": "vite build",
-    "preview": "vite preview --port 5173"
+    "preview": "vite preview --port 5173",
+    "test": "PW_TEST_CONNECT_WS_ENDPOINT=ws://127.0.0.1:3000/ playwright test --config playwright.config.ts",
+    "test:server": "docker run --add-host=hostmachine:host-gateway -p 3000:3000 --rm --init -d --user pwuser mcr.microsoft.com/playwright:v1.60.0-noble /bin/sh -c \"npx -y playwright@1.60.0 run-server --port 3000 --host 0.0.0.0\""
```
DoPri: I think, this should use `docker compose run` instead of plain `docker run`

.github/workflows/ci.yml
DoPri, line 1: nice!

DoPri (commented):
One suggestion, otherwise LGTM!

services/client/package.json
BjarneHa: `docker compose run` only starts services specified in the `docker-compose.yaml` if I understand correctly.
Replacing `docker run` with `docker compose run` yields an error because it cannot find the specified service.
Instead, I added the playwright server as an optional service to the `docker-compose.yaml`, which can be enabled with `docker compose --profile e2e up`. The `test:server` script has therefore been removed. What's your take on this approach?

DoPri: Sounds like a good compromise!

DoPri (approved):

services/client/playwright.config.ts
```diff
@@ -0,0 +1,73 @@
...
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: 'html',
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like `await page.goto('')`.
+     * The full stack (local and CI) is served via Traefik on port 80. */
+    baseURL: 'http://client',
```
ladu-tu, line 30: Could you explain where this URL comes from? The client app should be served via traefik on http://localhost

ladu-tu (commented):
One question.

services/client/playwright.config.ts
BjarneHa: Oh right, I forgot to ask about this. In the server code, I've seen requests to the reverse proxy (localhost instead of container names), however, substituting `http://localhost` as a `baseURL` yields `NS_ERROR_CONNECTION_REFUSED`. Am I doing something incorrectly or is this due to the fact that the reverse proxy is only reachable as `localhost` from outside the docker network?

BjarneHa: Note that the Playwright server, which sends the HTTP requests, is running inside our Docker network. Therefore `client` refers to the docker-internal domain name of the client container.

ladu-tu: I assume you are referring to `genai.base-url`? In that case, localhost is only used for contexts in which the genai service isn't needed (API spec generation etc.), otherwise it is always overriden with `GENAI_BASE_URL=http://genai:8000` from the .env.config file.

Your explanation with the docker network domain name makes sense, I didn't think about that!

PR #144, opened 2026-06-13
Task/clean up env files

### LINKED ISSUE #116: [Infra] Clean up env file split and fix docker-compose variable warnings
## Motivation

Running `docker compose up` on a fresh clone (without an `.env`) produces a wall of warnings:

```
WARN[0000] The "POSTGRES_DB" variable is not set. Defaulting to a blank string.
WARN[0000] The "POSTGRES_USER" variable is not set. Defaulting to a blank string.
...
```

This happens because Docker Compose automatically loads `.env` but we keep secrets in `.env` (gitignored) and non-secret config in `.env.config`. The `env_file: .env.config` directive loads the config, but Compose also tries to interpolate `${POSTGRES_DB}` etc. from `.env` at parse time -- before the env_file is applied -- and warns when they're missing there.

The naming `.env.config` is also non-standard and slightly confusing alongside `.env.example`.

## Proposed change

Pick one approach (preference: option A) and implement it:

**Option A -- `docker-compose.override.yml` with local dev defaults**
Add a `docker-compose.override.yml` (gitignored for secrets, but committed for safe defaults) that sets dummy-but-functional values for all variables that only matter locally. Compose merges overrides automatically, so `docker compose up` Just Works without warnings.

**Option B -- rename `.env.config` to `.env.defaults` and document the two-file pattern**
Cleaner naming. Requires updating every `env_file:` reference and the README.

**Option C -- provide a committed `.env.local.example` with safe local defaults**
Developers copy it to `.env` once after cloning. Simpler but requires a manual step.

Whichever approach is chosen:
- `docker compose up` on a fresh clone must produce no WARN lines about unset variables.
- The distinction between "committed config" and "gitignored secrets" must be clear from the file names alone.
- README and `.env.example` must be updated to match.

## Affects

- [x] Server (Spring)
- [ ] Client
- [x] GenAI
- [ ] API spec
- [x] CI / CD
- [x] Docs

## Definition of Done

- [ ] `docker compose up` on a fresh clone (no manual env setup) emits no WARN lines about unset variables
- [ ] The env file naming is self-explanatory without reading the README
- [ ] README explains the env file setup in one paragraph
- [ ] CI smoke test step works without extra setup (it currently relies on env vars set as GitHub vars/secrets)

ladu-tu:
## What does this PR do?
This PR resolves the issue concerning the Postgres environment variables being empty if no .env file is present. This leads to warnings if a local user has freshly cloned the repository without creating a local .env for development. CI and infra create their own .env files, so they are not affected.

This closes #116. I did not implement any of the approaches proposed by @DoPri, but rather simply added default fallback values in the docker-compose.yml itself and added documentation to make the .env.config/.env.example approach clearer. In my opinion, this is the simplest and cleanest fix for this issue, without introducing additional files.

## Type of change

- [X] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [X] Docs
- [ ] CI / infra

## Definition of Done

- [x] CI is green


DoPri (changes requested):
If we hardcode defaults in the docker-compose file (which is a fine solution imo), I don't see any reason why we would actually need a separate `.env.config` file.

At least to my knowledge, we didn't change any of the defaults defined in the `.env.config` in any of our deployments (Azure, Kubernetes, nor locally).
Having three places for configuration (`.env.config`, `.env`, and `docker-compose.yml`) seems overly and unnecessarily complicated to me.

TL;DR: I think, we should remove the `.env.config` altogether and move its config into the `docker-compose.yml`. User can still override the defaults in their local `.env` file. (We should also add comments there to tell the users which variables are available.)

ladu-tu:
@DoPri I agree that removing the .env.config would probably remove some clutter.

However, if we add documentation on which variables are overridable, would that not result in the same structure we already have in the .env.config? Additionally, providing the .env.config allows a user to more quickly set up their local development; they just need to copy the file to a local .env and edit it.

I would be fine with moving everything to the docs, but I don't see a great benefit from this. What do you think?

DoPri:
> would that not result in the same structure we already have in the .env.config?

No, since we would then only have configuration in two places instead of three, which is easier to understand and maintain.

> Additionally, providing the .env.config allows a user to more quickly set up their local development; they just need to copy the file to a local .env and edit it.

I don't get your point here. The users should already copy `.env.example` to `.env`. You say they should also copy `.env.config` to `.env`? But then they would have to merge these two files (`.env.example` and `.env.config`) into one local `.env`?? I'm confused. Are we really talking about the same thing here? I want to remove `.env.config`, not `.env.example`. Please clarify.

ladu-tu:
Sorry, now I myself was confused by all the config files.

How about we remove .env.config, but add the variables to .env.example (possibly commented out), so that it is clearly visible which variables may be overridden?

DoPri:
Great idea! That's a good compromise!

ladu-tu:
@DoPri Please check whether the changes I made to the Azure deployment are ok.

On another note, the duplication between docker-compose.yaml and docker-compose.azure.yaml is quite a headache to deal with and may lead to issues when changes are not applied to both files.
As they currently only differ regarding the build process, I wonder if it might be suitable to use [compose file merging](https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/#how-to-merge-multiple-compose-files)? What do you think about that? We could open an issue regarding it.

DoPri:
Yes, please create an issue for that.
I'll test and review this PR later today.

docker-compose.azure.yml
```diff
@@ -48,13 +53,18 @@ services:
...
     image: ghcr.io/aet-devops26/team-bnd/genai:latest
     container_name: genai
-    env_file:
-      - .env.config
+    environment:
+      GENAI_LOG_LEVEL: ${GENAI_LOG_LEVEL:-info}
+      LLM_PROVIDER: ${LLM_PROVIDER:-logos}
+      LLM_BASE_URL: ${LLM_BASE_URL:-https://logos.aet.cit.tum.de/v1}
+      LLM_MODEL: ${LLM_MODEL:-openai/gpt-oss-120b}
+      LLM_API_KEY: ${LLM_MODEL:-"lg-your-api-key"}
```
DoPri: Copy & Paste error here.

docker-compose.azure.yml
```diff
@@ -110,9 +120,10 @@ services:
     env_file:
       - .env
     environment:
-      POSTGRES_DB: ${POSTGRES_DB}
-      POSTGRES_USER: ${POSTGRES_USER}
-      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
+      PGPORT: ${POSTGRES_PORT:-5432}
```
DoPri, line 123: Why `PGPORT`and not `POSTGRES_PORT`?

docker-compose.yml
```diff
@@ -57,6 +62,12 @@ services:
     image: ghcr.io/aet-devops26/team-bnd/genai:latest
     build: services/genai/
     container_name: genai
+    environment:
+      GENAI_LOG_LEVEL: ${GENAI_LOG_LEVEL:-info}
+      LLM_PROVIDER: ${LLM_PROVIDER:-logos}
+      LLM_BASE_URL: ${LLM_BASE_URL:-https://logos.aet.cit.tum.de/v1}
+      LLM_MODEL: ${LLM_MODEL:-openai/gpt-oss-120b}
+      LLM_API_KEY: ${LLM_MODEL:-"lg-your-api-key"}
```
DoPri: Same Copy & Paste error as before.

docker-compose.yml
```diff
@@ -109,12 +120,11 @@ services:
   postgres-db:
     image: postgres:16-alpine
     container_name: postgres-db
-    env_file: .env.config
     environment:
-      POSTGRES_DB: ${POSTGRES_DB}
-      POSTGRES_USER: ${POSTGRES_USER}
-      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
-      PG_PORT: ${POSTGRES_PORT}
+      PGPORT: ${POSTGRES_PORT:-5432}
```
DoPri, line 124: Again, same question as before.

docker-compose.azure.yml
DoPri: Also, the quoting breaks inside the container things and should be avoided.

docker-compose.yml
DoPri: Also, the quoting breaks inside the container things and should be avoided.

DoPri (changes requested):

docker-compose.azure.yml
ladu-tu: Fixed.

ladu-tu: Fixed.

docker-compose.yml
ladu-tu: Fixed.

ladu-tu: Fixed.

docker-compose.azure.yml
ladu-tu: The only offically documented environment variable for setting the PostgreSQL port I found was `PGPORT` (cf. [postgres docs](https://www.postgresql.org/docs/current/app-postgres.html)). The `POSTGRES_*` variables we set are documented for the [docker image](https://hub.docker.com/_/postgres), but there is no port variable.

If you look closely at the official repo for the postgres images, [docker-entrypoint.sh](https://github.com/docker-library/postgres/blob/be0b7ac960bb393b00a31938b6d878397f482759/docker-entrypoint.sh) reads in the env variables, mapping e.g. `POSTGRES_USER` to `PGUSER`, but nothing like that is done for `POSTGRES_PORT`. It doesn't even appear anywhere in the repo.
Apparently, other people were also confused by that (cf. issues [here](https://github.com/docker-library/postgres/issues/1000), [here](https://github.com/docker-library/postgres/issues/702) and [here](https://github.com/docker-library/postgres/issues/196)). But the developers don't want to implement this for some reason. 

I also tested it locally, any custom port is only set correctly if we pass `PGPORT`; `POSTGRES_PORT` is silently ignored.

TL;DR: `PGPORT` is the correct environment variable.

docker-compose.yml
ladu-tu: Refer to my answer above.

DoPri (approved):

PR #148, opened 2026-06-17
Refactor: drop redundant docker-compose.azure.yml

### LINKED ISSUE #145: [Infra] Reduce docker-compose duplication using file merging
## Motivation

As discussed with @DoPri in PR #144, the duplication between `docker-compose.yml` and `docker-compose.azure.yml` is a maintenance burden. When changes are made to one file, they often need to be manually applied to the other, which can lead to:
- Inconsistencies between local dev and Azure deployment
- Bugs when changes are forgotten in one file
- Extra review effort on PRs

Currently, the two files only differ in:
- Image references (local uses `build:`, Azure uses pre-built `image:`)
- Environment file loading (`.env` vs `.env.config`)
- Minor configuration differences

## Proposed change

Use [Docker Compose file merging](https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/) to eliminate duplication:

### Option A: Base + Override pattern

```
docker-compose.yml          # Base config (shared)
docker-compose.override.yml # Local dev overrides (build contexts)
docker-compose.azure.yml    # Azure overrides (pre-built images)
```

Usage:
- Local dev: `docker compose up` (auto-merges override)
- Azure: `docker compose -f docker-compose.yml -f docker-compose.azure.yml up`

### Option B: Base + Profiles

Use [profiles](https://docs.docker.com/compose/how-tos/profiles/) to conditionally include build contexts:

```yaml
services:
  spring:
    image: ghcr.io/aet-devops26/team-bnd/spring:latest
    profiles: ["azure"]
    # ... shared config

  spring-dev:
    extends: spring
    build: services/spring/
    profiles: ["dev"]
```

### Recommended: Option A

Option A is simpler and matches existing Docker Compose conventions. The override pattern is well-documented and widely used.

### File structure after refactoring

```
docker-compose.yml          # All service definitions, no build contexts, uses ${IMAGE_TAG:-latest}
docker-compose.override.yml # Adds build: context for each service (for local dev)
docker-compose.azure.yml    # Sets specific image tags, env_file: .env (for Azure)
```

### CI/CD changes

Update `.github/workflows/ci.yml`:
- Smoke test: uses default merge (docker-compose.yml + override)
- Azure deploy: explicitly specify `-f docker-compose.yml -f docker-compose.azure.yml`

## Affects

- [x] Server (Spring)
- [x] Client
- [x] GenAI
- [ ] API spec
- [x] CI / CD
- [x] Docs
- [x] Infrastructure

## Definition of Done

- [ ] Single source of truth for service definitions
- [ ] Local dev works with `docker compose up`
- [ ] Azure deployment works with explicit file merge
- [ ] CI smoke test passes
- [ ] Azure deployment pipeline updated
- [ ] README documents the file structure

DoPri:
## What does this PR do?

Removes `docker-compose.azure.yml` and points the Azure VM deploy at the base `docker-compose.yml`. Closes #145.

The two compose files had drifted into being almost identical, the only thing the azure file added on top of the base was `env_file: .env` on `spring` and `postgres-db`. That turns out to be redundant: `docker compose` already auto-loads `./.env` and substitutes it into the existing `${VAR:-default}` blocks, so the base file on its own resolves every secret (Postgres, Keycloak, LLM key) once a `.env` sits next to it on the VM. The `build:` contexts in the base don't get in the way either, `docker compose pull` and `up -d` use the `image:` refs and never build (we don't pass `--build` on the VM).

So instead of the base + override split the issue suggested, this just drops the second file entirely. One compose file, nothing to keep in sync by hand.

Deploy job change: scp `docker-compose.yml` instead of the azure file, and run `docker compose pull && docker compose up -d` against the default file.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Docs
- [x] CI / infra

## Definition of Done

- [x] CI is green
- [x] If the API changed: `api/openapi.yaml` is updated and lint passes (n/a, no API change)
- [ ] Tests added or updated for the changed behaviour (n/a)
- [x] Docs updated where it matters (`infra/azure/README.md`)

## Notes for the reviewer

I verified locally that `docker compose -f docker-compose.yml config` with a `.env` present resolves all the secrets that the old azure file injected via `env_file`, the merged output was byte-identical to the old standalone azure config. So this is behaviour-preserving for the VM.

The one thing I couldn't test from my machine is the actual `docker compose pull && up -d` against GHCR on the VM (needs registry auth + the running VM). The image refs are unchanged, so it should behave exactly as before, but the first deploy on main is the real confirmation, worth a quick check that the VM still comes up green after merge.

The k8s/stud deploy path is untouched, it never used the azure compose file.


BjarneHa (approved):
LGTM!

PR #153, opened 2026-06-18
feat(client): use typescript for development

### LINKED ISSUE #120: [Bug] Client Dockerfile uses Node 18 (EOL) while CI builds with Node 24
## What happened

The client's `Dockerfile` uses `node:18-alpine` as the build base image, but Node 18 reached end of life in April 2025 and no longer receives security updates. Meanwhile, `ci.yml` builds the client with Node 24. The two environments don't match, which means a build that passes CI could in theory behave differently in the Docker image.

## Steps to reproduce

1. Look at `services/client/Dockerfile` line 1: `FROM node:18-alpine AS build`
2. Look at `.github/workflows/ci.yml`, `build-client` job: `node-version: "24"`

## Expected vs actual

**Expected:** the Node version used in the Dockerfile matches what CI uses, and is a supported LTS release.

**Actual:** Dockerfile is two major versions behind CI and is EOL.

## Environment

- Component: client
- Branch / commit: main
- Local or k8s: both (Dockerfile affects local `docker compose up` and the published image)

## Logs / screenshots

No runtime failure yet, but `docker scout` or Trivy will flag the base image as having known CVEs.

## Fix

Update the `Dockerfile` build stage to `node:24-alpine` (or pin to a specific patch like `node:24-alpine3.21` for reproducibility). No code changes should be needed -- the build is just `npm ci && npm run build`, which is Node-version-agnostic for this project.

Also add an `.nvmrc` or `engines` field in `package.json` so local devs and CI are all on the same version without having to read the Dockerfile.

BjarneHa:
## What does this PR do?

This PR contains the basic typescript setup for the client. Closes #120.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] CI / infra

## Definition of Done

- [ ] CI is green


services/client/package.json
```diff
@@ -15,6 +15,9 @@
   "devDependencies": {
     "@playwright/test": "^1.60.0",
     "@types/node": "^25.9.0",
+    "@types/react": "^18",
+    "@types/react-dom": "^18",
+    "typescript": "^5",
```
DoPri: Why not use the latest version?

services/client/package.json
```diff
@@ -15,6 +15,9 @@
   "devDependencies": {
     "@playwright/test": "^1.60.0",
     "@types/node": "^25.9.0",
+    "@types/react": "^18",
+    "@types/react-dom": "^18",
```
DoPri: Why not use the latest version?

services/client/package.json
```diff
@@ -15,6 +15,9 @@
   "devDependencies": {
     "@playwright/test": "^1.60.0",
     "@types/node": "^25.9.0",
+    "@types/react": "^18",
```
DoPri: Why not use the latest version?

services/client/package.json
```diff
@@ -15,6 +15,9 @@
...
+    "@types/react-dom": "^18",
+    "typescript": "^5",
     "vite": "^5.0.0"
```
DoPri: Why not use the latest version?

DoPri (changes requested):

DoPri (commented):
Please consider adding an [engines](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines) entry to `packages.json`. Otherwise, LGTM.

DoPri (approved):

PR #155, opened 2026-06-19
infra(helm): fix failing k8s deployment due to incorrect keycloak config

BjarneHa:
## What does this PR do?

This PR configures the Helm chart to enable the Keycloak backchannel via `extraArgs` instead of `hostnameBackchannel` (which is no longer supported).
It also imports `realm.json` into the k8s deployment and increases the max. allowed memory consumption of Keycloak.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] CI / infra

## Definition of Done

- [x] CI is green


infra/k8s/alexandria/files/oidc
```diff
@@ -0,0 +1 @@
+../../../oidc
```
ladu-tu, line 1: Would it be possible to turn this into a project root relative (somewhat absolute) path? This looks like it could be error prone.

infra/k8s/alexandria/values.yaml
```diff
@@ -70,10 +74,10 @@ keycloak:
   resources:
     requests:
       cpu: 250m
-      memory: 256Mi
+      memory: 512Mi
```
ladu-tu, line 77: Was there an out-of-memory issue for the database or why did you increase this value?

BjarneHa: Yes. Keycloak would sometimes be OOMKilled.

infra/k8s/alexandria/files/oidc
BjarneHa: I wouldn't know how to create such a symlink.

PR #161, opened 2026-06-23
Infra: add healthchecks to every service and gate startup on them

### LINKED ISSUE #147: [Feature] Add healthchecks to every service and gate startup on them
## Motivation

Right now only Traefik has a healthcheck in `docker-compose.yml`. Spring and GenAI define a `HEALTHCHECK` in their Dockerfiles, but client, Keycloak, and Postgres have none, and `depends_on` uses the plain list form so nothing actually waits for a dependency to be healthy. Spring depends on Keycloak, Postgres, and GenAI, but it'll start the moment those containers exist, not when they're ready, which gives us flaky first-boot failures. Keycloak in particular takes a while to import the realm and is the worst offender.

## Proposed change

Add a healthcheck to every service in `docker-compose.yml` (and `docker-compose.azure.yml`) and switch the `depends_on` entries that matter to the `condition: service_healthy` long form.

Per service:
- client (nginx): add a compose healthcheck hitting the nginx root, and add a `HEALTHCHECK` to `services/client/Dockerfile` (it currently has none).
- spring: add a compose healthcheck against `/actuator/health` (Dockerfile already has one, mirror it in compose so `depends_on` can gate on it).
- genai: add a compose healthcheck against `/genai/health` (Dockerfile already has one).
- keycloak: this is the important one. The image has no curl/wget, so a naive healthcheck won't work. Use Keycloak's own health endpoint properly -- e.g. enable `KC_HEALTH_ENABLED=true` and probe the management `/health/ready` endpoint, or use the bash `/dev/tcp` trick the community uses for this image. Give it a generous `start_period` since realm import is slow.
- postgres-db: add a `pg_isready` healthcheck.

Then update `depends_on`: Spring should wait for `postgres-db`, `keycloak`, and `genai` to be `service_healthy`, not just present.

## Affects

- [x] Server (Spring)
- [x] Client
- [x] GenAI
- [ ] API spec
- [x] CI / CD
- [ ] Docs

## Definition of Done

- [ ] Every service in `docker-compose.yml` has a working `healthcheck`
- [ ] `services/client/Dockerfile` has a `HEALTHCHECK`
- [ ] Keycloak healthcheck actually reflects readiness (realm imported, health endpoint green), not just "container started"
- [ ] Postgres uses `pg_isready`
- [ ] Spring's `depends_on` uses `condition: service_healthy` for postgres-db, keycloak, and genai
- [ ] `docker-compose.azure.yml` kept in sync with the same healthchecks
- [ ] `docker compose up` shows all services reaching `healthy` and Spring only starts serving once its deps are healthy

DoPri:
## What does this PR do?

Adds a healthcheck to every service in `docker-compose.yml` and switches the dependency edges that matter to `condition: service_healthy`, so Spring no longer starts before Postgres, Keycloak, and GenAI are actually ready. Also cleans up the CI now that `docker compose up --wait` gates on those healthchecks.

Closes #147.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] CI / infra

## Definition of Done

- [x] CI is green

## Details

Healthchecks:

- client: added a `HEALTHCHECK` to `services/client/Dockerfile` (it had none) plus a compose healthcheck hitting the nginx root.
- spring: compose healthcheck against `/actuator/health/readiness` (mirrors the Dockerfile so `depends_on` can gate on it).
- genai: compose healthcheck against `/genai/health`.
- keycloak: `KC_HEALTH_ENABLED=true` and probe the management health endpoint. The image has no curl/wget, so the check opens `/dev/tcp/localhost/9000` from bash and looks for an HTTP 200. The endpoint sits under `KC_HTTP_RELATIVE_PATH`, so it's `/auth/health/ready`, not `/health/ready`. `start_period` is 180s because cold JVM boot + realm import ran anywhere from 60s to 90s+ on my machine.
- postgres-db: `pg_isready`.

depends_on:

- spring waits for postgres-db, keycloak, and genai to be `service_healthy`.
- client waits for spring to be `service_healthy`.
- Dropped the `depends_on: traefik` entries. Traefik discovers containers through the Docker socket; the services don't need it up to be healthy and it doesn't need them, so gating on it was backwards.

CI (`.github/workflows/ci.yml`):

- compose-smoke: replaced the three `sleep`-based retry loops with single assertion curls. `--wait` already blocks until everything is healthy, so the polling was pointless.
- e2e: removed the `Wait for Keycloak Realm` polling step. Keycloak's healthcheck only goes green once the realm is imported, so `--wait` covers it.
- Bumped `--wait-timeout` from 180s to 360s. A cold build with Keycloak's slow boot pushed total startup to ~277s locally, so 180s would have been flaky.

## Notes for the reviewer

I found a pre-existing bug while doing this: the old compose-smoke genai check curled `/genai/health`, which Traefik doesn't route (only `/docs`, `/redoc`, `/openapi.json` are exposed), so it always 404'd. The old retry loop swallowed it (looped 5x, no `exit 1`), so it silently always passed. The new hard-assertion curl exposed it, so I pointed it at `/docs` instead, which Traefik actually serves.

Verified locally: clean `docker compose up --build --wait` brings all six services to healthy in the right order (deps first, then spring, then client) in ~277s, and the smoke curls all pass (`/hello` -> "Hello World!", `/` -> 200, `/docs` -> 200).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -388,18 +388,10 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: Bring Stack Up
-        # --wait blocks until every service is healthy (or the timeout hits),
-        # so the checks below are plain assertions, not polling loops.
+        # --wait blocks until every service reports healthy (or the timeout
+        # hits), so a successful up is the smoke test. The per-service
+        # healthchecks in docker-compose.yml are what actually gets asserted.
         run: docker compose up -d --build --wait --wait-timeout 360
-
-      - name: Check Spring Service
-        run: curl -fsS http://localhost/hello | grep -q "Hello World"
-
-      - name: Check Client Service
-        run: curl -fsS http://localhost/ > /dev/null
-
-      - name: Check GenAI Service
-        run: curl -fsS http://localhost/docs > /dev/null
 
       - name: Show Logs on Failure
         if: failure()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -388,37 +388,18 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: Bring Stack Up
-        run: docker compose up -d --build --wait --wait-timeout 180
+        # --wait blocks until every service is healthy (or the timeout hits),
+        # so the checks below are plain assertions, not polling loops.
+        run: docker compose up -d --build --wait --wait-timeout 360
 
       - name: Check Spring Service
-        run: |
-          for i in 1 2 3 4 5 6 7 8 9 10; do
-            if curl -fsS http://localhost/hello | grep -q "Hello World"; then
-              echo "spring ok"; break
-            fi
-            echo "waiting for spring service... ($i/10)"
-            sleep 3
-          done
+        run: curl -fsS http://localhost/hello | grep -q "Hello World"
 
       - name: Check Client Service
-        run: |
-          for i in 1 2 3 4 5; do
-            if curl -fsS http://localhost/ > /dev/null; then
-              echo "client ok"; break
-            fi
-            echo "waiting for client... ($i/5)"
-            sleep 2
-          done
+        run: curl -fsS http://localhost/ > /dev/null
 
       - name: Check GenAI Service
-        run: |
-          for i in 1 2 3 4 5; do
-            if curl -fsS http://localhost/genai/health > /dev/null; then
-              echo "genai ok"; break
-            fi
-            echo "waiting for genai... ($i/5)"
-            sleep 2
-          done
+        run: curl -fsS http://localhost/docs > /dev/null
 
       - name: Show Logs on Failure
         if: failure()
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -388,37 +388,18 @@ jobs:
...
       - name: Check Spring Service
-        run: |
-          for i in 1 2 3 4 5 6 7 8 9 10; do
-            if curl -fsS http://localhost/hello | grep -q "Hello World"; then
-              echo "spring ok"; break
-            fi
-            echo "waiting for spring service... ($i/10)"
-            sleep 3
-          done
+        run: curl -fsS http://localhost/hello | grep -q "Hello World"
```
ladu-tu: Considering that we now have fully functional healthchecks, do we even need these "Check XXX Service" steps anymore? This would allow us to restrict some further endpoints, e.g. `hello` not being accessible from the outside anymore. Alternatively, is there a way to execute this inside the docker network?

.github/workflows/ci.yml
```diff
@@ -388,37 +388,18 @@ jobs:
...
       - name: Check Client Service
-        run: |
-          for i in 1 2 3 4 5; do
-            if curl -fsS http://localhost/ > /dev/null; then
-              echo "client ok"; break
-            fi
-            echo "waiting for client... ($i/5)"
-            sleep 2
-          done
+        run: curl -fsS http://localhost/ > /dev/null
```
ladu-tu: Same as above.

.github/workflows/ci.yml
```diff
@@ -388,37 +388,18 @@ jobs:
...
       - name: Check GenAI Service
-        run: |
-          for i in 1 2 3 4 5; do
-            if curl -fsS http://localhost/genai/health > /dev/null; then
-              echo "genai ok"; break
-            fi
-            echo "waiting for genai... ($i/5)"
-            sleep 2
-          done
+        run: curl -fsS http://localhost/docs > /dev/null
```
ladu-tu: Same as above.

DoPri:
Good call, I just deleted the now redundant checks and also de-depublished the Spring `/hello` endpoint.

ladu-tu (approved):
Thanks!

PR #162, opened 2026-06-23
Feat: Object Storage Setup

ladu-tu:
## What does this PR do?

Previously, all documents were stored as BLOBs (binary large objects) in the PostgreSQL database. There are multiple reasons as to why this is not best practice (cf. issue #133). This PR therefore switches to SeaweedFS, a S3 compatible storage and adds a dedicated storage service. Additionally, the Spring service is adapted to upload/download documents from the file storage container instead of storing them in the database. This closes #133, closes #134 and closes #135.

## Type of change

- [ ] Bug fix
- [X] New feature
- [X] Refactor / cleanup
- [x] Docs
- [X] CI / infra

## Definition of Done

- [x] CI is green
- [X] If the API changed: `api/openapi.yaml` is updated and lint passes
- [X] If a service was added or restructured: docker-compose still comes up cleanly (`docker compose up`)
- [X] Tests added or updated for the changed behaviour
- [X] Docs updated where it matters (README, ADRs, comments)

## Notes for the reviewer

The PostgreSQL database schemas have changed, you will have to remove the corresponding docker volume before spinning up the containers, otherwise the Spring service will fail.

**Important**: @DoPri As agreed upon in our meeting, I changed the communication between the Spring service and the GenAI service to no longer pass the full document, but instead the objectKey of the corresponding document. The GenAI client can then get the respective document from the file storage service itself. Refer to #136 for further information.

@BjarneHa As you have already set up the helm deployment, it would be nice if you could handle issue #137.

Issue #137, opened 2026-06-13
[Infra] Update Helm chart with file storage service


ladu-tu:
## Parent Issue

#133 - Switch file storage from PostgreSQL BLOBs to S3 compatible object storage

## Proposed change

Add file storage to Kubernetes deployment via Helm chart.

## Affects

- [x] Infrastructure
- [ ] Server (Spring)
- [ ] GenAI

## Definition of Done

- [ ] Helm dependency added
- [ ] PersistentVolume configured
- [ ] Default bucket created on startup
- [ ] Spring ConfigMap includes file storage settings
- [ ] GenAI ConfigMap includes file storage settings
- [ ] Secrets configured for credentials

Issue #133, opened 2026-06-13
[Server] Switch file storage from PostgreSQL BLOBs to S3 compatible object storage


ladu-tu:
## Motivation

Currently, uploaded files are stored as BLOBs in PostgreSQL via the `FileContent` entity. This approach has several drawbacks:
- Database size grows rapidly with file uploads
- Inefficient for large files (PDFs, images)
- Backup/restore becomes slower and more complex
- GenAI service cannot directly access files (requires Spring to proxy)

MinIO provides S3-compatible object storage that scales independently and allows both Spring and GenAI to access files directly. However, it has been recently deprecated (cf. [MinIO repository](https://github.com/minio/minio)). Therefore, an alternative storage service should be chosen.

## Proposed change

Replace PostgreSQL BLOB storage with S3 compatible object storage:

| Aspect | Current | Proposed |
|--------|---------|----------|
| Storage | PostgreSQL `FileContent` table | S3-compatible buckets |
| Access | Spring only (via JPA) | Spring + GenAI (via S3 API) |
| Scalability | Limited by DB size | Independent scaling |
| Backup | Part of DB dump | Separate object backup |
| API | Custom | Standard S3 API |

## Sub-issues

- [ ] #134 - Add file storage service to docker-compose
- [ ] #135 - Refactor Spring DocumentService to use file storage service
- [ ] #136 - Update GenAI service to read files from file storage
- [ ] #137 - Update Helm chart with file storage service

## Affects

- [x] Server (Spring)
- [ ] Client
- [x] GenAI
- [ ] API spec
- [x] CI / CD
- [ ] Docs
- [x] Infrastructure

## Definition of Done

- [x] File storage container running in docker-compose
- [ ] Spring uploads files to file storage instead of PostgreSQL
- [ ] GenAI can access uploaded files via S3 API
- [ ] `FileContent` entity removed or deprecated
- [ ] Helm chart updated for Kubernetes deployment
- [ ] Existing functionality preserved (upload, retrieve, delete)

PR #167, opened 2026-06-25
Feat: GenAI reads documents from object storage

### LINKED ISSUE #136: [GenAI] Update GenAI service to read files from file storage
## Parent Issue

#133 - Switch file storage from PostgreSQL BLOBs to S3 compatible object storage

## Proposed change

Update GenAI service to read document files from the file storage service using the S3 API. This includes adding the corresponding dependencies, creating an S3 client (e.g. `boto3`). The `/genai/summarize` and `/genai/extract` endpoitns have to be modified to accept object keys or file paths instead of text content, retrieving the corresponding documents from the file storage.

## Affects

- [x] GenAI
- [ ] Server (Spring)
- [ ] Infrastructure

## Definition of Done

- [ ] boto3 added to dependencies
- [ ] S3 client configured
- [ ] Endpoints can read files by object key
- [ ] PDF text extraction works
- [ ] Backward compatible (still accepts text content)

### LINKED ISSUE #170: [Security] Verify genai Python CVE fixes and rebuild the image
## Motivation

The trivy filesystem and genai image scans from the guest lecture flagged two HIGH severity CVEs in transitive Python deps of the GenAI service:

- langsmith 0.8.14 -- GHSA-f4xh-w4cj-qxq8 (HIGH), fixed in 0.8.18
- starlette 1.2.1 -- CVE-2026-54283 (HIGH), fixed in 1.3.1, plus CVE-2026-54282 (LOW), fixed in 1.3.0

Both come in transitively (langsmith via langchain-openai, starlette via fastapi). The scan ran against an older snapshot. Our current `services/genai/uv.lock` already resolves langsmith to 0.9.2 and starlette to 1.3.1, so both HIGH CVEs look already fixed by the recent genai dependency bump. This is a verify-and-rebuild, not a fresh fix.

## Proposed change

Confirm the lockfile is actually clean and that the published image picks it up:

- Double-check `services/genai/uv.lock` has langsmith >= 0.8.18 and starlette >= 1.3.1.
- Rebuild the genai image so the registry tag no longer ships the vulnerable versions.
- Run trivy against the freshly built image locally to confirm the two HIGH findings are gone.

If for some reason the lock still pins old versions in CI, run `uv lock --upgrade-package langsmith --upgrade-package starlette` and commit the result.

## Affects

- [ ] Server (Spring)
- [ ] Client
- [x] GenAI
- [ ] API spec
- [x] CI / CD
- [ ] Docs

## Definition of Done

- [ ] `uv.lock` confirmed at langsmith >= 0.8.18 and starlette >= 1.3.1
- [ ] genai image rebuilt and pushed
- [ ] `trivy image` on the new genai tag reports no HIGH findings for langsmith or starlette
- [ ] genai tests still pass after any lock change

DoPri:
## What does this PR do?

Updates the GenAI service to read documents straight from the object storage instead of getting text content in the request body. The `/genai/summarize`, `/genai/extract` and `/genai/ask` endpoints now take object keys; the service downloads each file from the SeaweedFS bucket (the same one Spring uploads to), extracts the text, and feeds that to the LLM. Plain text is decoded as UTF-8, PDFs go through pypdf. This closes #136.

This is the GenAI half of the storage migration from #133. The Spring client (from #162) already sends `objectKey` / `objectKeys` and expects `sourceObjectKeys` back, so this just makes the two services actually line up.

This PR should only be merged after #162.

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Refactor / cleanup
- [x] Docs
- [ ] CI / infra

## Definition of Done

- [x] CI is green

## Notes for the reviewer

A few things worth knowing:

- I dropped the "still accepts text content" point from the issue's DoD. Spring doesn't send raw text anymore, so keeping that path would just be dead, untested branching.
- New `app/storage.py` holds the boto3 client and `fetch_text`. It uses path-style addressing because SeaweedFS needs it (same as Spring's `forcePathStyle(true)`), and reads the same `S3_*` env vars Spring uses so both point at one bucket.
- Error handling: missing key -> 404, non-text/non-PDF -> 415, empty document -> 422. For `ask`, unreadable keys are skipped so one bad document doesn't kill the whole request.
- I verified the storage path end to end against a real SeaweedFS container (text object and a PDF object), plus the 404/422 paths through the running service. The actual LLM call is unchanged and still covered by mocked tests.
- @BjarneHa the k8s chart has no object storage yet, so I left the Helm side alone. Could you wire up the SeaweedFS deployment + the `S3_*` config for spring and genai there as part of the storage work in #162 / #137? Right now GenAI would start without S3 config on the cluster.

## Summary by CodeRabbit

* **New Features**
  * GenAI requests now work with stored document object keys instead of raw text, including ask, extract, and summarize flows.
  * Responses now return source object keys for easier traceability.
  * The service can now read UTF-8 text and PDF documents from object storage.

* **Bug Fixes**
  * Added clearer handling for missing, unreadable, and unsupported documents.
  * Ask requests now skip unreadable files so one bad object does not fail the whole request.

* **Documentation**
  * Updated API and service docs to match the new object-key based behavior.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -5,7 +5,7 @@ info:
   license:
     name: MIT
     identifier: MIT
-  version: 1.3.0
+  version: 0.2.0
 servers:
   - url: /
     description: Current server
@@ -29,6 +29,43 @@ tags:
     name: ai
     x-displayName: ai
 paths:
+  /api/v1/knowledgebase/documents:
+    get:
+      tags:
+        - KnowledgeBase
+      summary: List all documents
+      operationId: listDocuments
+      responses:
+        '200':
+          description: Documents retrieved
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Document'
+      security:
+        - bearerAuth: []
+    post:
+      tags:
+        - KnowledgeBase
+      summary: Create a new document with text content
+      operationId: createDocument
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateDocumentRequest'
+        required: true
+      responses:
+        '201':
+          description: Document created
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Document'
+      security:
+        - bearerAuth: []
   /api/v1/knowledgebase/documents/{id}/tags:
     post:
       tags:
@@ -81,27 +118,6 @@ paths:
                 $ref: '#/components/schemas/Document'
         '400':
           description: Invalid file
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/Document'
-      security:
-        - bearerAuth: []
-  /api/v1/knowledgebase/documents/create:
-    post:
-      tags:
-        - KnowledgeBase
-      summary: Create a new document with text content
-      operationId: createDocument
-      requestBody:
-        content:
-          application/json:
-            schema:
-              $ref: '#/components/schemas/CreateDocumentRequest'
-        required: true
-      responses:
-        '201':
-          description: Document created
           content:
             application/json:
               schema:
@@ -207,23 +223,6 @@ paths:
                   $ref: '#/components/schemas/QAInteraction'
       security:
         - bearerAuth: []
-  /api/v1/knowledgebase/documents:
-    get:
-      tags:
-        - KnowledgeBase
-      summary: List all documents
-      operationId: listDocuments
-      responses:
-        '200':
-          description: Documents retrieved
-          content:
-            application/json:
-              schema:
-                type: array
-                items:
-                  $ref: '#/components/schemas/Document'
-      security:
-        - bearerAuth: []
   /api/v1/knowledgebase/documents/{id}:
     get:
       tags:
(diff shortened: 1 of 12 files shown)


api/openapi.yaml
```diff
@@ -5,7 +5,7 @@ info:
   license:
     name: MIT
     identifier: MIT
-  version: 0.3.0
+  version: 0.2.0
```
ladu-tu: Wrong API version. If you want to bump it, please update the Spring side and GenAI to the same version, otherwise these changes get lost during API merging.

services/genai/README.md
```diff
@@ -29,6 +33,18 @@ The service is configured entirely via environment variables. Copy `.env.example
 | `LLM_API_KEY`     | _(required for logos/openai)_     | API key (`lg-...` for Logos, `sk-...` for OAI) |
 | `OLLAMA_BASE_URL` | `http://localhost:11434`          | Ollama server URL (provider=ollama only)       |
 
+## Object storage configuration
```
ladu-tu: This section should not be in the GenAI documentation, as it belongs to multiple services. Please move this to a dedicated docs file.

services/genai/README.md
```diff
@@ -29,6 +33,18 @@ The service is configured entirely via environment variables. Copy `.env.example
...
+
+The service reads documents from the S3-compatible object storage. These match the variables the Spring service uses so both point at the same bucket.
+
+| Variable        | Default                  | Description                         |
+| --------------- | ------------------------ | ----------------------------------- |
+| `S3_ENDPOINT`   | `http://s3-storage:8333` | S3 gateway URL                      |
+| `S3_REGION`     | `eu-central-1`           | Region label (SeaweedFS ignores it) |
```
ladu-tu: Unnecessary comment at this place imo.

ladu-tu:
Once #162 is merged, please rebase to main.

DoPri:
I'll remove the documentation here. @ladu-tu please add it in your #162

ladu-tu:
The OpenAPI spec shouldn't be modified manually.

Please update the SERVICE_VERSION on the GenAI side and `services/spring/app/src/main/java/com/alexandria/app/config/OpenApiConfig.java` on the Spring side instead.

Issue #175, opened 2026-06-26
[Docs] Add a SECURITY.md with a vulnerability reporting policy


DoPri:
## Motivation

The lecture's "Security through the Toolchain" section made the point that a security researcher who finds a hole in your repo needs a documented way to tell you (SECURITY.md, security.txt, a contact). We already have a CODEOWNERS file covering repo access, and the codeowners check passed, but there's no SECURITY.md, so there's no stated channel for reporting a vulnerability.

This is a small, standard addition that rounds out the toolchain-hardening side of the lecture.

## Proposed change

Add a `SECURITY.md` (GitHub picks it up from the repo root or `.github/`) with:

- which versions / branches we consider supported (just `main` for a course project is fine)
- how to report a vulnerability (a contact, e.g. the team's course email or a private GitHub security advisory)
- rough expectation on response time

Keep it short and honest about this being a university project. GitHub's "Report a vulnerability" private advisory flow is the simplest channel to point at, so enabling private vulnerability reporting in repo settings and referencing it here is the path of least resistance.

## Affects

- [ ] Server (Spring)
- [ ] Client
- [ ] GenAI
- [ ] API spec
- [ ] CI / CD
- [x] Docs

## Definition of Done

- [ ] `SECURITY.md` exists with supported versions, a reporting channel, and a response expectation
- [ ] GitHub shows the security policy in the repo's Security tab
- [ ] (optional) private vulnerability reporting enabled in repo settings

PR #190, opened 2026-06-28
Docs: Fix typo, add typos ignore config, add SECURITY.md

### LINKED ISSUE #174: [Docs] Fix the real typo and add a typos ignore for the false positives
## Motivation

The typos scanner reported 58 hits. Only one is a real typo we should fix; the other 57 are false positives that need a typos config so they stop showing up.

- Real: `docs/System_Overview.md:25` -- "natual language questions" should be "natural".
- False positive (56 hits): `autheticatorFlow` in `infra/oidc/realm.json`. This is misspelled, but it's a Keycloak realm-export field name -- Keycloak itself spells it that way. Renaming it breaks the realm import, so we can't touch it.
- False positive (1 hit): `infra/k8s/README.md:7` -- typos thinks `AKS` should be `ASK`. AKS is Azure Kubernetes Service, leave it.

## Proposed change

- Fix the one real typo in `docs/System_Overview.md` (natual -> natural).
- Add a typos config (`.typos.toml`) at the repo root that ignores the Keycloak field name and the AKS acronym, so the scan goes clean without us pretending the realm field is correct English. Roughly:

```toml
[default.extend-words]
AKS = "AKS"

[default.extend-identifiers]
autheticatorFlow = "autheticatorFlow"
```

(confirm the exact keys against how typos is invoked in CI before committing)

## Affects

- [ ] Server (Spring)
- [ ] Client
- [ ] GenAI
- [ ] API spec
- [ ] CI / CD
- [x] Docs

## Definition of Done

- [ ] "natual" fixed to "natural" in `docs/System_Overview.md`
- [ ] `.typos.toml` ignores `autheticatorFlow` and `AKS`
- [ ] typos run is clean
- [ ] realm.json left untouched (import still works)

DoPri:
## What does this PR do?

Closes #174 and #175. Also normalizes `gradlew.bat` line endings.

- #174: fixes the one real typo flagged by the typos scanner (`natual` -> `natural` in `docs/System_Overview.md`) and adds a `.typos.toml` at the repo root so the two false-positive classes stop showing up. `autheticatorFlow` is a Keycloak realm-export field that is misspelled upstream and renaming it breaks the realm import, and `AKS` is Azure Kubernetes Service. Scan goes from 57 errors to 0; realm.json untouched.
- #175: adds `SECURITY.md` at the repo root with supported versions (only `main`, ends with the course), a reporting channel (GitHub private vulnerability reporting as primary, CODEOWNERS via Artemis as fallback), a rough response expectation, and scope.
- Drive-by: `services/spring/app/gradlew.bat` was committed to history with CRLF stored in the blob, even though `.gitattributes` declares `*.bat text eol=crlf` (which normalizes the blob to LF and only checks out as CRLF). This branch stages the normalization so the blob matches what gitattributes wants. Working-tree file is unchanged (still CRLF on checkout), no behavior change.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [x] Docs
- [ ] CI / infra

## Definition of Done

- [ ] CI is green
- [ ] If the API changed: `api/openapi.yaml` is updated and lint passes -- n/a, no API change
- [ ] If a service was added or restructured: docker-compose still comes up cleanly (`docker compose up`) -- n/a
- [ ] Tests added or updated for the changed behaviour -- n/a, docs and line-ending normalization only


ladu-tu (approved):
LGTM

Issue #179, opened 2026-06-26
[Security] Extend CodeQL to the genai (Python) and client (TypeScript) code


DoPri:
## Motivation

We run CodeQL for static analysis, but the matrix in `.github/workflows/codeql.yml` only lists `java-kotlin`, so only the Spring service gets scanned. The genai Python service and the client TypeScript are invisible to SAST right now, even though CodeQL supports both and the matrix is set up to make adding a language a one-line change. The lecture framed SAST (semgrep/CodeQL) as a baseline part of the toolchain, and "More Languages and Technologies to Secure" was one of its core points: each language is its own attack surface.

## Proposed change

- Add `python` and `javascript-typescript` to the CodeQL language matrix.
- Python and JS/TS use CodeQL's no-build analysis, so they don't need the manual gradle build step the Java path uses. Guard the Java-only build step so it only runs for `java-kotlin`, and let the other languages autobuild.
- Fix the job name expression so it renders for all three languages, not just Java (right now `name:` only has a branch for `java-kotlin`).
- Widen the workflow's `paths:` triggers so changes under `services/genai/**` and `services/client/**` actually kick off a scan (currently it only triggers on `services/spring/**`).

## Affects

- [x] Server (Spring)
- [x] Client
- [x] GenAI
- [ ] API spec
- [x] CI / CD
- [ ] Docs

## Definition of Done

- [ ] CodeQL matrix covers java-kotlin, python, and javascript-typescript
- [ ] The manual build step runs only for Java; Python and TS analyze without it
- [ ] Job name renders correctly for every language
- [ ] `paths:` triggers cover genai and client changes
- [ ] All three CodeQL jobs pass on a PR

Issue #176, opened 2026-06-26
[Security] Fix dependabot coverage so every ecosystem actually gets updates


DoPri:
## Motivation

`.github/dependabot.yml` only watches three things: github-actions at `/`, gradle at `/services/spring/app`, and docker at `/services/spring`. That leaves most of the repo's dependencies with no automated updates at all, which is the gap the lecture's whole "keep dependencies fresh" section warns about. It's also why the genai CVEs in #170 had to be bumped by hand: dependabot was never watching the genai Python deps in the first place.

What's currently missing:

- npm at `/services/client` -- the client deps (there's a `package-lock.json`), completely untracked
- uv at `/services/genai` -- the genai Python deps (`pyproject.toml` + `uv.lock`), untracked. This is exactly where the langsmith/starlette HIGH CVEs lived.
- docker at `/services/genai` -- base images `ghcr.io/astral-sh/uv:python3.14-alpine` and `python:3.14-alpine`
- docker at `/services/client` -- base images `node:24-alpine` and `nginx:stable-alpine`
- terraform at `/infra/azure/terraform` -- Azure provider and module versions

One thing to keep in mind: dependabot's docker updater only bumps the `FROM` tag/digest, it doesn't patch OS packages inside the image. But bumping the digest is what triggers the rebuild that pulls patched packages, so it still matters for the spring image CVEs in #171.

## Proposed change

Extend `.github/dependabot.yml` to cover every ecosystem. dependabot supports `uv` as its own ecosystem (distinct from `pip`) and reads `pyproject.toml` + `uv.lock`, so we can track genai properly. New entries:

- `npm` -> `/services/client`
- `uv` -> `/services/genai`
- `docker` -> add `/services/genai` and `/services/client` (use the `directories:` list form to fold the docker entries together rather than repeating the block)
- `terraform` -> `/infra/azure/terraform`

While we're in here, add the `cooldown:` block #173 calls for to each ecosystem (e.g. `semver-patch-days: 3`, `semver-minor-days: 7`) so we're not adopting brand-new releases the same day they ship. Keep the existing grouping so we don't drown in PRs, and set sane `open-pull-requests-limit` values per ecosystem.

## Affects

- [x] Server (Spring)
- [x] Client
- [x] GenAI
- [ ] API spec
- [x] CI / CD
- [ ] Docs

## Definition of Done

[...]

PR #192, opened 2026-06-28
CI: Harden workflows, extend dependabot, widen CodeQL matrix

### LINKED ISSUE #173: [Security] Harden GitHub Actions workflows per zizmor findings
## Motivation

zizmor (the GitHub Actions security linter from the lecture) found 67 issues across `ci.yml` and `codeql.yml`: 37 errors, 27 notes, 3 warnings. These map onto the OWASP Top 10 CI/CD risks the lecture walked through, mainly Dependency Chain Abuse, Poisoned Pipeline Execution, and credential hygiene. Worth doing because the workflows are the highest-value attack surface in the repo and these are the patterns the guest lecturers specifically called out.

Breakdown:

- unpinned-uses (36, error): every `uses:` is pinned to a tag (`actions/checkout@v6`, `dorny/paths-filter@v4`, `hadolint/hadolint-action@v3.3.0`, etc.) rather than a full commit SHA. Tags are mutable, so a hijacked tag could run arbitrary code in our pipeline. zizmor wants SHA pins.
- template-injection (15, mostly note + 1 error): `${{ ... }}` expanded straight into `run:` shell blocks in the Azure deploy job (around `ci.yml:545-565`). The `env.AZURE_*` ones are low risk since we set those values, but `${{ github.actor }}` going into a shell string (the `docker login` step) is the classic injection vector.
- artipacked (13, note): `actions/checkout` steps don't set `persist-credentials: false`, so the GITHUB_TOKEN stays in `.git/config` for every later step in the job.
- dependabot-cooldown (3, warning): `.github/dependabot.yml` has no cooldown, so we'd pull brand-new dependency versions immediately. The lecture recommended a cool-down period (e.g. 3 days) to let supply-chain attacks surface before we adopt a release.

## Proposed change

Tackle in order of value:

1. template-injection: replace inline `${{ github.actor }}` (and ideally the other expansions) in `run:` blocks with `env:` entries on the step, then reference the shell variable. This is the only real injection risk, do it first.
2. artipacked: add `persist-credentials: false` to every `actions/checkout` step that doesn't need the token to persist (which is all of them in this repo, since we don't push back from those jobs).
3. dependabot-cooldown: add a `cooldown:` block to each ecosystem in `.github/dependabot.yml` (e.g. 3 days for minor/patch). This is handled together with the dependabot coverage gaps in #176, so coordinate with that one rather than editing the file twice.
4. unpinned-uses: pin third-party actions to commit SHAs. For first-party `actions/*` and `github/codeql-action/*` this is lower risk; decide whether to SHA-pin everything (cleanest, but noisier dependabot PRs) or just the third-party ones (`dorny/paths-filter`, `hadolint/hadolint-action`). Whatever we pick, keep dependabot's github-actions updater on so the SHAs still get bumped.

Going forward: run zizmor as a CI step on workflow changes so new workflows don't reintroduce these. Adding it to the existing security jobs would make the lecture's point land in our own pipeline.

## Affects

- [ ] Server (Spring)
- [ ] Client
- [ ] GenAI
- [ ] API spec
- [x] CI / CD
- [ ] Docs

## Definition of Done

- [ ] `${{ github.actor }}` and other risky expansions moved out of `run:` blocks into step `env:`
- [ ] All `actions/checkout` steps set `persist-credentials: false` unless they provably need the token
- [ ] dependabot cooldown handled (see #176)
- [ ] Third-party actions pinned to commit SHAs (decision on first-party pinning documented either way)
- [ ] zizmor added as a CI step so regressions get caught
- [ ] CI still green after the changes

DoPri:
## What does this PR do?

Closes #173, #176, #179.

- #173: harden workflows per zizmor. Move `${{ }}` expansions out of `run:` blocks into step `env:`, add `persist-credentials: false` to every `actions/checkout`, SHA-pin third-party actions, add a `zizmor` job as a regression guard.
- #176: extend `dependabot.yml` to npm, uv, docker (genai + client), and terraform. Add `cooldown` (`default-days: 7`, patch 3, minor 7) to every ecosystem.
- #179: add `python` and `javascript-typescript` to the CodeQL matrix, guard the Gradle build to `java-kotlin`, widen `paths:` triggers to genai and client.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] CI / infra

## Definition of Done

- [x] CI is green
- [ ] Docs updated where it matters (README, ADRs, comments)

## Notes for the reviewer

- Verified locally: `actionlint` clean, `zizmor` (default persona) 0 findings, `dependabot.yml` valid per schemastore dependabot-2.0 schema.
- First-party actions stay on tag refs by policy; see `.github/zizmor.yml`.
- Also adds typos to `.pre-commit-config.yml` (Kudos @ladu-tu #191)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,10 @@ updates:
     schedule:
       interval: weekly
     open-pull-requests-limit: 5
+    cooldown:
+      default-days: 7
+      semver-patch-days: 3
+      semver-minor-days: 7
     groups:
       gha-minor:
         update-types: [minor, patch]
@@ -14,6 +18,10 @@ updates:
     schedule:
       interval: weekly
     open-pull-requests-limit: 5
+    cooldown:
+      default-days: 7
+      semver-patch-days: 3
+      semver-minor-days: 7
     groups:
       spring-boot:
         patterns:
@@ -23,7 +31,50 @@ updates:
         update-types: [minor, patch]
 
   - package-ecosystem: docker
-    directory: "/services/spring"
+    directories:
+      - "/services/spring"
+      - "/services/genai"
+      - "/services/client"
     schedule:
       interval: weekly
     open-pull-requests-limit: 3
+    cooldown:
+      default-days: 7
+      semver-patch-days: 3
+      semver-minor-days: 7
+
+  - package-ecosystem: npm
+    directory: "/services/client"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 5
+    cooldown:
+      default-days: 7
+      semver-patch-days: 3
+      semver-minor-days: 7
+    groups:
+      npm-minor:
+        update-types: [minor, patch]
+
+  - package-ecosystem: uv
+    directory: "/services/genai"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 5
+    cooldown:
+      default-days: 7
+      semver-patch-days: 3
+      semver-minor-days: 7
+    groups:
+      uv-minor:
+        update-types: [minor, patch]
+
+  - package-ecosystem: terraform
+    directory: "/infra/azure/terraform"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 3
+    cooldown:
+      default-days: 7
+      semver-patch-days: 3
+      semver-minor-days: 7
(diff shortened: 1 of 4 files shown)


ladu-tu (approved):
LGTM.

Issue #172, opened 2026-06-26
[Security] Harden container securityContext in the Helm templates


DoPri:
## Motivation

kics flagged 99 misconfigurations across `docker-compose.yml` and the Helm templates under `infra/k8s/alexandria/templates/`. Most are notes/warnings that don't apply to a student project (LimitRange, ResourceQuota, AppArmor profiles, image digest pinning in templates), but a few are worth doing because they're cheap and match the container-hardening part of the lecture.

The real one: our deployments set a pod-level `securityContext` (runAsNonRoot, runAsUser) but no container-level one. kics flags "Privilege Escalation Allowed" on the spring, genai, and client containers because `allowPrivilegeEscalation` isn't set to false, and "Root Container Not Mounted Read-only" because there's no `readOnlyRootFilesystem`.

The error-level kics findings that are NOT worth acting on, for the record:
- "Docker Socket Mounted In Container" at `docker-compose.yml:259` -- this is Traefik mounting the socket read-only, which is how Traefik discovers containers. Expected, leave it.
- "Passwords And Secrets - Generic Password" in `docker-compose.yml` -- these are the `${VAR:-default}` local-dev fallback values (e.g. `localpgpassword`), documented as safe defaults in `.env.example`. Not real secrets.
- client `runAsNonRoot: false` is deliberate: the nginx image needs to bind port 80.

## Proposed change

Add a container-level `securityContext` to the spring, genai, and client containers in their Helm templates:

```yaml
securityContext:
  allowPrivilegeEscalation: false
  readOnlyRootFilesystem: true
  capabilities:
    drop:
      - ALL
```

Notes:
- `readOnlyRootFilesystem: true` may need an `emptyDir` mount for any path a service writes to at runtime (temp dirs, caches). Check each service and add volume mounts where needed, or drop `readOnlyRootFilesystem` for the one that genuinely needs a writable root and document why.
- The client (nginx) needs a writable path for its cache and pid file, so it'll need the emptyDir treatment if we keep readOnlyRootFilesystem there.

## Affects

- [x] Server (Spring)
- [x] Client
- [x] GenAI
- [ ] API spec
- [x] CI / CD
- [ ] Docs

## Definition of Done

- [ ] spring, genai, and client containers have a container-level securityContext with `allowPrivilegeEscalation: false` and dropped capabilities
- [ ] `readOnlyRootFilesystem` applied where the service can run with it, with emptyDir mounts for required writable paths
[...]

PR #202, opened 2026-06-29
CI: Add gitleaks and kics scanners to CI and pre-commit

### LINKED ISSUE #178: [Security] Run gitleaks and kics in CI so the lecture's findings can't regress
## Motivation

The guest lecture ran a whole battery of scanners against us (gitleaks, trivy, hadolint, dockle, kics, zizmor, typos). Our own CI already runs trivy, hadolint, typos, and CodeQL, but gitleaks and kics only ever ran on the lecturers' side. So the secret-leak and IaC-misconfiguration classes of finding have no guard in our pipeline: once we fix them (#172 for kics), nothing stops them coming back.

gitleaks in particular the lecture wanted as early as possible, ideally a pre-commit hook, so a secret never even gets committed. The "attackers run these tools too" slide is the argument: catch it before they do.

(zizmor is the third missing scanner, but that's already tracked in #173, so this issue is just gitleaks and kics.)

## Proposed change

- Add a gitleaks step to CI on PRs and pushes. Also add it to `.pre-commit-config.yaml` so it runs locally before a commit lands, which is where the lecture said it has the most value.
- Add a kics step to CI that scans the IaC: `docker-compose.yml`, the Helm templates under `infra/k8s/`, and the Terraform under `infra/azure/`. Upload its SARIF so findings show up in the Security tab next to trivy and CodeQL.
- Decide per scanner whether it gates (fails the build) or just reports. gitleaks should gate, a leaked secret is a hard stop. kics is noisier, so it's reasonable to start it as non-gating (report-only) with a documented allowlist for the false positives we already identified in #172 (Traefik socket, local-dev compose defaults), then tighten it to gating once the noise is filtered.

## Affects

- [ ] Server (Spring)
- [ ] Client
- [ ] GenAI
- [ ] API spec
- [x] CI / CD
- [ ] Docs

## Definition of Done

- [ ] gitleaks runs in CI and as a pre-commit hook, and gates on a real secret
- [ ] kics runs in CI over compose, Helm, and Terraform, with SARIF uploaded
- [ ] kics false positives from #172 are allowlisted with a documented reason
- [ ] A planted test secret makes the gitleaks step fail (verify it actually works)

DoPri:
## What does this PR do?

Closes #178. Adds gitleaks (secret scanning) and kics (IaC misconfiguration scanning) to CI, plus a gitleaks pre-commit hook. These are the two scanners from the W09 guest lecture that we didn't have in our own pipeline yet.

**gitleaks**:

- New CI job that scans the full git history (`fetch-depth: 0`) on every PR and push.
- Runs gitleaks via the official Docker image (ghcr.io/gitleaks/gitleaks:8.30.1) instead of `gitleaks/gitleaks-action@v3`, which would require a license key for org repos. Consistent with the `gitleaks-docker` pre-commit hook.
- Gates the build: a real secret is a hard stop (exit code 1).
- SARIF uploaded to the Security tab.
- Pre-commit hook (`gitleaks-docker`) added to `.pre-commit-config.yaml` so secrets get caught before they're even committed. Uses the Docker image variant for consistency with the existing `hadolint-docker` hook.
- `.gitleaksignore` added to suppress two false positives from git history: a hardcoded test JWT secret (`"test-secret-key-that-is-at-least-32-bytes-long-for-hmac"`) in a `JwtServiceTest.java` that no longer exists in the tree but still shows up because gitleaks scans full history.

**kics**:

- New CI job scanning `docker-compose.yml`, Helm templates (`infra/k8s/alexandria`), and Terraform (`infra/azure/terraform`).
- Starts non-gating (`ignore_on_exit: results`): reports findings without failing the build.
- SARIF uploaded to the Security tab alongside trivy, gitleaks, and CodeQL.
- False positives from #172 are excluded via `.kics.yml` (`config_path` in the workflow), which is also runnable locally with `kics scan --config .kics.yml`:
  - `d6355c88-...` (Docker Socket Mounted In Container): Traefik mounts the socket read-only to discover containers. Expected.
  - `a88baa34-...` (Passwords And Secrets): docker-compose.yml uses `${VAR:-default}` fallback values for local dev. Not real secrets.
- kics is CI-only, not in pre-commit. gitleaks is a fast hard gate (sub-second, catches secrets before commit), so it belongs in pre-commit. kics is slower and noisier -- even with the allowlist it produces findings that are better reviewed as SARIF in the Security tab than blocked at commit time.

Both jobs are wired into `ci-summary`.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] CI / infra

## Definition of Done

- [x] CI is green

## Notes for the reviewer

kics is non-gating for now. Once #172 lands and the remaining noise is filtered, we can tighten it to gating by switching `ignore_on_exit: results` to `fail_on: high,medium`.

The kics SARIF file path is `kics-results/results.sarif` (kics outputs to `output_path` with a default filename).

## Summary by CodeRabbit

* **New Features**
  * Added automated secret scanning and infrastructure-as-code security checks to the CI pipeline.
  * Expanded path-based change detection so infrastructure updates trigger the appropriate security validation jobs.
* **Bug Fixes**
  * Suppressed a few known historical security findings to reduce noise during scans.
* **Chores**
  * Updated CI reporting so the new security checks are included in the overall pass/fail result.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
       client: ${{ steps.filter.outputs.client }}
       genai: ${{ steps.filter.outputs.genai }}
       docker: ${{ steps.filter.outputs.docker }}
+      infra: ${{ steps.filter.outputs.infra }}
       workflows: ${{ steps.filter.outputs.workflows }}
       docs: ${{ steps.filter.outputs.docs }}
     steps:
@@ -78,6 +79,11 @@ jobs:
               - '**/Dockerfile.*'
               - 'docker-compose.yml'
               - 'docker-compose.*.yml'
+            infra:
+              - 'docker-compose.yml'
+              - 'docker-compose.*.yml'
+              - 'infra/k8s/**'
+              - 'infra/azure/**'
             workflows:
               - '.github/workflows/**'
             docs:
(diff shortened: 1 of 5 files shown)


ladu-tu (approved):
LGTM if you apply the CI change as proposed by CodeRabbit.

PR #242, opened 2026-07-08
GenAI: Add content-based document tagging endpoint

### LINKED ISSUE #187: [GenAI] Auto-assign content-based tags on document upload
## Motivation

User_Stories #8 and #9 say Alexandria should automatically tag documents by content so the knowledge base gets categorised without manual effort, and show those tags in the document view. We extract entities already, but there's no tagging: nothing assigns topical tags, and the document view has nothing to show. This is a self-contained GenAI feature that makes the categorisation and tag-filtering stories (#12) actually possible.

## Proposed change

- Add a tagging capability to the genai service that takes document text (extracted from object storage, same as the other endpoints since #167) and returns a small set of topical tags. Use the LLM with a structured-output prompt, similar to how entity extraction already works, capped to a handful of tags with a controlled-ish vocabulary so we don't get 30 one-off tags per document.
- Expose it either as its own endpoint (e.g. `POST /genai/tag`) or fold it into the existing add-document processing alongside summarize/extract. Match whatever pattern the post-#167 flow uses, and update the API spec.
- Spring persists the tags with the document (there's already a search/tags area in the schema) and returns them so the client can show and filter by them. Coordinate the contract with the Spring knowledge-base flow.
- Tests with the LLM mocked, README updated.

If we want tag-similarity later (e.g. clustering tags), the embedding work from the RAG epic could help, but this issue does not depend on it; a structured LLM call is enough.

## Affects

- [x] Server (Spring)
- [x] Client
- [x] GenAI
- [x] API spec
- [ ] CI / CD
- [x] Docs

## Definition of Done

- [ ] genai returns a bounded set of content-based tags for a document
- [ ] Tagging runs as part of the document-add flow, API spec updated
- [ ] Tags are persisted by Spring and returned to the client
- [ ] Tests cover the tagging path with the LLM mocked
- [ ] README documents the tagging behaviour

DoPri:
## What does this PR do?

Adds a content-based tagging capability to the GenAI service. New `POST /genai/tag` takes an object key, pulls the document text from object storage (same flow as summarize/extract), and asks the LLM via structured output for a small set of topical tags describing what the document is about.

Closes #187.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] CI / infra

## Definition of Done

- [x] CI is green

## Summary by CodeRabbit

* **New Features**
  * Added content-based tagging via `POST /genai/tag`, returning topical `tags` and `modelUsed`.
  * Documented tagging in the GenAI service README, including `knownTags` support to bias reuse of existing labels.
* **Bug Fixes**
  * Improved tagging output by normalizing labels, deduplicating, capping the number of tags, and returning an empty list when no clear topics are found.
  * Added/verified consistent 404/415/422 handling for the tagging flow.
* **Chores**
  * Bumped API/service versions (including OpenAPI to `2.2.0` and GenAI to `2.1.0`) and updated OpenAPI metadata.
  * Added automated tests for the new tagging endpoint and tag generation logic.


api/openapi.yaml
```diff
@@ -1307,6 +1340,30 @@ components:
...
+    GenAiTagRequest:
+      properties:
+        objectKey:
+          title: Objectkey
+          type: string
+      required:
+        - objectKey
+      title: GenAiTagRequest
+      type: object
+    GenAiTagResponse:
```
ladu-tu, line 1358: The IndexResponse also returns the ObjectKey again, why doesn't the TagResponse?

api/openapi.yaml
```diff
@@ -1307,6 +1340,30 @@ components:
...
+      properties:
+        modelUsed:
```
ladu-tu, line 1360: The IndexResponse returns `embeddingModel`, all other endpoints `modelUsed`. Why that inconsistency?

services/genai/app/tag.py
```diff
@@ -0,0 +1,60 @@
+"""Content-based tagging using LangChain structured output."""
+
+from langchain_core.prompts import ChatPromptTemplate
+from pydantic import BaseModel, Field
+
+from app.llm import get_llm, get_model_name
+
+# Keep the number of tags per document small so the knowledge base stays
+# browsable instead of drowning in one-off labels.
```
ladu-tu, line 11: To avoid one-off tags, it would probably be best if the a list of all current tags was passed from the Spring service to the GenAI service, so the latter one can reuse labels. Should we implement that as GenAI requesting Spring or Spring passing them in the TagRequest?

services/genai/tests/test_tag.py
```diff
@@ -0,0 +1,63 @@
...
+    with patch("app.tag.get_llm", return_value=mock_llm), patch("app.tag.get_model_name", return_value="openai/gpt-oss-120b"):
+        from app.tag import generate_tags
+
+        tags, model = generate_tags("A paper about NLP research methods.")
+
+    assert model == "openai/gpt-oss-120b"
+    assert tags == ["machine learning", "nlp", "research"]
+
+
+def test_generate_tags_lowercases_trims_and_deduplicates():
```
ladu-tu, line 29: Do Tags really have to be lowercase? I would be fine with e.g. `Finance` too instead of `finance`.

services/genai/README.md
```diff
@@ -12,12 +12,15 @@ The processing endpoints take object keys, not raw text. The service downloads t
...
 | POST   | `/genai/extract`           | Extract named entities from the document at `objectKey`      |
+| POST   | `/genai/tag`               | Assign content-based topical tags to the document at `objectKey` |
 | POST   | `/genai/index`             | Chunk, embed, and index the document at `objectKey`          |
 | DELETE | `/genai/index/{objectKey}` | Remove a document's chunks from the index                    |
 | POST   | `/genai/ask`               | Answer a question from indexed chunks of the given documents |
 | GET    | `/genai/metrics`           | Prometheus metrics (in-network scraping only)                |
 
-`summarize`, `extract`, and `index` take `{"objectKey": "..."}`. A key that is missing returns 404 (415 if the object is neither UTF-8 text nor a PDF, 422 if it has no extractable text).
+`summarize`, `extract`, `tag`, and `index` take `{"objectKey": "..."}`. A key that is missing returns 404 (415 if the object is neither UTF-8 text nor a PDF, 422 if it has no extractable text).
+
```
ladu-tu, line 22: Seems like the longer description of `summarize`, `extract` and `index` are missing.

ladu-tu (approved):

ladu-tu (changes requested):
Sorry, I misclicked. Some questions before approving.

api/openapi.yaml
DoPri: IndexResponse echoes objectKey because it's a write op. summarize and extract don't echo it either; they derive content from the stored object. tag is also a read/derive op, so it's consistent with summarize/extract, not index. No change here. If you want it changed, please open a separate issue.

api/openapi.yaml
DoPri: Again, tag is consistent with summarize, extract, and ask. They all return `modelUsed` (the chat model). index returns `embeddingModel` because it's the one endpoint that produces embeddings and reports which embedder generated the vectors. This is an intentional exception and it predates this PR. tag uses the chat model, so `modelUsed` is correct.

services/genai/app/tag.py
DoPri: Agreed. My suggestion: Spring passes the user's existing tags in the request (keeps the dependency one-way and GenAI stays free of Spring's API). `GenAiTagRequest` now has an optional `knownTags` field that biases the prompt toward reuse. Spring wiring (populating `knownTags` from the tags table) is a follow-up.

services/genai/tests/test_tag.py
DoPri: Keeping lowercase for deduplication. `Finance` and `finance` collapse to one tag instead of fragmenting the vocabulary across documents. Display casing is a client concern. No change here.

services/genai/README.md
DoPri: Fair. Will fix.

api/openapi.yaml
ladu-tu: I don't really understand that rationale, but I'm fine with it.

ladu-tu (approved):

ladu-tu:
@DoPri Could you please also create an issue for the Spring side implementation?

Issue #11, opened 2026-05-05
[Client] Simple Hello World Application


ladu-tu:
This feature implements a basic "Hello World" react app. Additionally, this task includes setting up a preliminary Dockerfile to run the app in a containerized manner.

**Related user stories:**

**Acceptance criteria:**
- [x] A basic react app is set up
- [ ] The Dockerfile allows to build and run the app locally

**Tasks**
- [ ] Set up Dockerfile
- [ ] Add client code

PR #243, opened 2026-07-08
GenAI: Add semantic search endpoint over the knowledge base

### LINKED ISSUE #188: [GenAI] Add a semantic search endpoint over the knowledge base
## Motivation

User_Stories #10-13 cover search: find documents by keyword, see why they matched, filter by tags and entities. Keyword search can live entirely on the Spring/Postgres side, but semantic search (find documents about a concept even when the exact words differ) wants vector similarity, which is exactly the embedding infrastructure the RAG epic builds. Once chunks are embedded and in Weaviate, a search endpoint is a small addition that reuses all of it.

## Proposed change

- Add a search capability to the genai service: embed the query with the same embedding config as the RAG epic (#180), run a similarity search in Weaviate, and return the matching documents ranked by relevance, deduplicated from chunk hits up to the document level, with enough info (matched chunk snippet, score) for the client to show why each result matched (#11).
- Expose as `POST /genai/search` (or similar) and update the API spec. Spring calls it and either merges the results with, or presents them alongside, its own keyword and tag/entity filters; decide how semantic and keyword results combine with the Spring side.
- This depends on the RAG embedding + Weaviate work landing first (#180), specifically the embedding module and the indexed chunks. It's tracked separately so the RAG epic stays focused, but it shouldn't start until that infrastructure exists.
- Tests with Weaviate and the embedder mocked, README updated.

## Affects

- [x] Server (Spring)
- [x] Client
- [x] GenAI
- [x] API spec
- [ ] CI / CD
- [x] Docs

## Definition of Done

- [ ] `/genai/search` embeds a query and returns documents ranked by semantic similarity
- [ ] Results carry match context (snippet/score) so the client can show why a document matched
- [ ] Spring integrates semantic results with its keyword/tag/entity filtering
- [ ] Depends-on relationship to the RAG epic noted; not started before that infra lands
- [ ] Tests cover the search path; README updated

DoPri:
## What does this PR do?

Adds a semantic search capability to the GenAI service. New `POST /genai/search` embeds the query with the same embedding config as the RAG pipeline, runs a vector similarity search in Weaviate over the indexed chunks, and rolls the chunk hits up to the document level so each document appears once, ranked by its closest-matching chunk. Every result carries a snippet and a relevance score so the client can show why a document matched (#11).

Closes #188.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] CI / infra

## Definition of Done

- [x] CI is green

## Notes for the reviewer

- This branch is based on `feat/187-genai-tagging`. Merge #242 first, then this targets `main`; otherwise the diff will also show the tagging commit.

## Summary by CodeRabbit

* **New Features**
  * Added `POST /genai/search` for semantic document search with `query`, scoped `objectKeys`, and optional `limit`.
  * Returns ranked results including `score`, `snippet`, `chunkIndex`, and the `embeddingModel`.

* **Bug Fixes**
  * Scoped searches with no document keys now return an empty result set (no broader/unscoped search).

* **Documentation**
  * Updated API documentation/contract and service docs to describe the new search request/response.

* **Tests**
  * Added API and unit test coverage for validation, ranking, scoring, and snippet behavior.

### DIFF
--- a/services/genai/app/main.py
+++ b/services/genai/app/main.py
@@ -100,7 +100,7 @@ class GenAiAskResponse(BaseModel):
 
 
 class GenAiSearchRequest(BaseModel):
-    query: str
+    query: str = Field(max_length=1500)
     objectKeys: list[str]
     limit: int = Field(default=10, ge=1, le=50)
 
--- a/services/genai/app/main.py
+++ b/services/genai/app/main.py
@@ -3,7 +3,7 @@ from contextlib import asynccontextmanager
 from fastapi import FastAPI, HTTPException
 from fastapi.responses import PlainTextResponse
 from prometheus_fastapi_instrumentator import Instrumentator
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, field_validator
 
 from app.embeddings import chunk_text, embed_chunks, get_embedding_model_name
 from app.extract import extract_entities
@@ -88,9 +88,17 @@ class GenAiTagResponse(BaseModel):
     modelUsed: str
 
 
+def _reject_blank(value: str) -> str:
+    if not value.strip():
+        raise ValueError("must not be blank")
+    return value
+
+
 class GenAiAskRequest(BaseModel):
     question: str
     objectKeys: list[str]
+
+    _validate_question = field_validator("question")(_reject_blank)
 
 
 class GenAiAskResponse(BaseModel):
@@ -103,6 +111,8 @@ class GenAiSearchRequest(BaseModel):
     query: str = Field(max_length=1500)
     objectKeys: list[str]
     limit: int = Field(default=10, ge=1, le=50)
+
+    _validate_query = field_validator("query")(_reject_blank)
 
 
 class GenAiSearchResult(BaseModel):
@@ -209,9 +219,6 @@ def ask(request: GenAiAskRequest) -> GenAiAskResponse:
     The question is embedded and matched against the indexed chunks scoped to the
     requested object keys; the answer cites the documents its chunks came from.
     """
-    if not request.question.strip():
-        raise HTTPException(status_code=422, detail="question must not be empty")
-
     answer, source_keys, model = answer_question(request.question, request.objectKeys)
     return GenAiAskResponse(answer=answer, sourceObjectKeys=source_keys, modelUsed=model)
 
@@ -224,9 +231,6 @@ def search(request: GenAiSearchRequest) -> GenAiSearchResponse:
     requested object keys; chunk hits are rolled up to one entry per document,
     each carrying the closest chunk's snippet and a relevance score.
     """
-    if not request.query.strip():
-        raise HTTPException(status_code=422, detail="query must not be empty")
-
     results, model = search_documents(request.query, request.objectKeys, request.limit)
     return GenAiSearchResponse(
         results=[GenAiSearchResult(objectKey=r["object_key"], score=r["score"], snippet=r["snippet"], chunkIndex=r["chunk_index"]) for r in results],
--- a/services/genai/app/main.py
+++ b/services/genai/app/main.py
@@ -3,18 +3,19 @@ from contextlib import asynccontextmanager
 from fastapi import FastAPI, HTTPException
 from fastapi.responses import PlainTextResponse
 from prometheus_fastapi_instrumentator import Instrumentator
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from app.embeddings import chunk_text, embed_chunks, get_embedding_model_name
 from app.extract import extract_entities
 from app.qa import answer_question
+from app.search import search_documents
 from app.storage import ObjectNotFoundError, UnsupportedFileError, fetch_text
 from app.summarize import summarize
 from app.tag import generate_tags
 from app.vectorstore import close_client, delete_document, index_chunks
 
 SERVICE_NAME = "alexandria-genai"
-SERVICE_VERSION = "2.1.0"
+SERVICE_VERSION = "2.2.0"
 
 
 @asynccontextmanager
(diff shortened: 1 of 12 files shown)


BjarneHa:
FYI: Notes for reviewers references the issue, not the PR.

BjarneHa (approved):
LGTM

ladu-tu:
Please add a (sub)issue for the Spring side implementaion, as in #243. Also, this PR doesn't close #188 as that specifies `Spring integrates semantic results with its keyword/tag/entity filtering` in the DoD. I would suggest moving that item to the Spring side issue.

services/genai/app/search.py
```diff
@@ -0,0 +1,80 @@
...
+
+def _score(distance: float | None) -> float:
+    # Weaviate returns cosine distance (0 = identical). Flip it into a similarity
+    # score where higher means more relevant, which is friendlier for the client.
+    if distance is None:
+        return 0.0
+    return round(1.0 - distance, 4)
+
+
+def _best_per_document(hits: list[dict], limit: int) -> list[dict]:
```
ladu-tu: SEARCH_CANDIDATES chunk hits are fetched, then deduplicated by `object_key`. If a document has many chunks that all rank in the top 50 (SEARCH_CANDIDATES), only that one document is returned even when limit=10 was asked for. Consider fetching more candidates and iterate until the limit is filled or use a Weaviate grouped query (group_by=object_key).

services/genai/app/search.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def _snippet(text: str) -> str:
+    collapsed = " ".join(text.split())
+    if len(collapsed) <= _SNIPPET_MAX_CHARS:
+        return collapsed
+    return collapsed[:_SNIPPET_MAX_CHARS].rstrip() + "..."
+
+
+def _score(distance: float | None) -> float:
```
ladu-tu, line 28: This score can be negative, as `_score = 1.0 - distance` where the Weaviate cosine distance can go up to 2.0 ([Reference](https://docs.weaviate.io/weaviate/config-refs/distances)). The API specifies "higher = more relevant" but does not give a range, thus clients may assume [0,1]. Either map this to [0,1] or document the range.

services/genai/app/main.py
```diff
@@ -87,6 +99,24 @@ class GenAiAskResponse(BaseModel):
     modelUsed: str
 
 
+class GenAiSearchRequest(BaseModel):
```
ladu-tu, line 110: Just as an optional hardening (this would also apply to some other endpoints): Currently, there is no length limit on the query, thus this could be used to inject overly long queries that would be forwarded to the upstream provider. I think a limit of ~1000 - 1500 chars should be sufficient, ofc that would need to be documented. That could also be shown to the user on the client side (e.g. input field for search query with note x/1500 chars).

services/genai/app/search.py
```diff
@@ -0,0 +1,80 @@
...
+
+def _candidates() -> int:
+    return int_env("SEARCH_CANDIDATES", _DEFAULT_CANDIDATES, minimum=1)
+
+
+def _snippet(text: str) -> str:
+    collapsed = " ".join(text.split())
+    if len(collapsed) <= _SNIPPET_MAX_CHARS:
+        return collapsed
+    return collapsed[:_SNIPPET_MAX_CHARS].rstrip() + "..."
```
ladu-tu: This may cut off words in the middle, as it is char-based. Consider cutting off at the last whitespace before.

ladu-tu (changes requested):

services/genai/app/search.py
DoPri: Good catch. Switched to Weaviate's group_by on object_key instead of the manual dedup, so the server does it directly. Dropped SEARCH_CANDIDATES entirely since it's not needed anymore.

services/genai/app/search.py
DoPri: Fixed, clamped with max(0.0, 1.0 - distance) so the score stays in 0, 1. Documented the range in the module docstring and the README.

services/genai/app/main.py
DoPri: Added max_length=1500 on the query field. Shows up as maxLength in the OpenAPI schema so the client can use it for an input limit too.

services/genai/app/search.py
DoPri: Done, now cuts at the last whitespace before the 300-char limit.

services/genai/app/vectorstore.py
```diff
@@ -150,6 +150,42 @@ def search(query_vector: list[float], object_keys: list[str], limit: int) -> lis
     ]
 
 
+def search_grouped(query_vector: list[float], object_keys: list[str], limit: int) -> list[dict]:
+    """Return one nearest chunk per distinct object_key (document-level results).
+
+    Uses Weaviate's group_by on ``object_key`` so each document appears once,
+    ranked by its closest chunk. ``number_of_groups`` caps the documents returned
+    and ``objects_per_group=1`` keeps only that closest chunk. An empty
```
ladu-tu, line 161: "An empty ``object_keys`` list searches the whole collection." But `search_documents` breaks earlier if no objectKeys are provided. So either update this comment or call `search_grouped` even if no keys are provided.

ladu-tu (changes requested):

services/genai/app/vectorstore.py
DoPri: Good catch. This also applies to the existing search(). Fixed both: empty object_keys now returns [] instead of searching the whole collection, so neither function can leak across users if a caller forgets to scope. Docstrings updated to match.

ladu-tu: Thanks!

ladu-tu (approved):
Please address coderabbitai's review, otherwise LGTM.

PR #261, opened 2026-07-09
Task: Automatically generate REST client for Spring service from OpenAPI spec

ladu-tu:
## What does this PR do?

This PR addresses and closes #251. The `knowledgebase` and `qa` services no longer use a raw REST client, but instead one generated based on the OpenAPI spec.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] CI / infra

## Definition of Done

- [x] CI is green

## Summary by CodeRabbit

* **New Features**
  * Added/expanded knowledgebase document management: reprocessing for summaries/entities, document summary and entity retrieval, and metadata updates.
  * Added document tagging plus GenAI search and GenAI tagging.
  * Added QA endpoints for asking questions and managing QA history, including history deletion.
  * Added user preference updates and internal cleanup/purge support.
* **Bug Fixes**
  * Improved API documentation/response definitions for “summary not yet available” (HTTP 204).
* **Tests**
  * Added integration coverage for GenAI extraction/type mapping behavior.


services/spring/openapi.yaml
DoPri, line 1: This is a verbatim copy of `api/openapi.yaml`. This should be a symlink at the very least.

.pre-commit-config.yaml
```diff
@@ -73,6 +73,12 @@ repos:
         language: script
         files: '^(services/genai/.*\.py|services/spring/.*\.java)$'
         pass_filenames: false
+      - id: sync-openapi-spring
+        name: OpenAPI - Move spec to Spring build context
+        entry: bash -c 'cp api/openapi.yaml services/spring/openapi.yaml'
```
DoPri: See above. Just copying verbatim is not very smart imo.

.pre-commit-config.yaml
```diff
@@ -87,9 +93,10 @@ repos:
         files: '^api/openapi\.yaml$'
         pass_filenames: false
       - id: client-generate-api
-        name: client-generate-api
-        entry: npm run generate:api --prefix services/client
-        language: system
+        name: OpenAPI - Generate Client
+        entry: openapi-typescript
+        language: node
+        additional_dependencies: ["openapi-typescript@7.13.0"]
```
DoPri, line 93: Why are these changes here?

services/spring/knowledgebase-service/build.gradle
```diff
@@ -11,6 +12,37 @@ openApi {
 	}
 }
 
+def genaiGeneratedDir = layout.buildDirectory.dir("generated/openapi").get().asFile
```
DoPri: This block is just a 1:1 copy from the other `build.gradle` files. Should be just one common block in the root `build.gradle`

DoPri (changes requested):

services/spring/openapi.yaml
ladu-tu: Symlinks are not followed and loaded for docker contexts (cf. [this discussion](https://github.com/moby/moby/issues/1676#issuecomment-23878459)). The YAML is however needed during the build to generate the client. The possibility of both copies going out of sync is also minimal to (if properly used) non-existent, as the hook copies the file on each API spec rebuild.

Do you have any ideas on how to improve it? Otherwise I would suggest leaving it as-is.

.pre-commit-config.yaml
ladu-tu: See above.

.pre-commit-config.yaml
ladu-tu: Because otherwise the pre-commit hook failed for me. It was expecting openapi-typescript to be installed on the machine. I fixed it to be platform-independent.

services/spring/openapi.yaml
DoPri: Agreed symlinks won't work in a Docker context. But the answer isn't to commit `api/openapi.yaml` a second time. The same file checked in twice is the anti-pattern we should avoid.

It only exists because the build context is `services/spring/`, so the Dockerfile can't reach the real spec. My proposal: Widen the context to a dir that contains both `api/` and `services/spring/` (repo root, or add `api/` as a named build context), point gradle at the single source (`inputSpec = "$rootDir/../../api/openapi.yaml"`) and `COPY api/openapi.yaml` from the real path in the Dockerfile. Then there is only one source of truth.

Also, regarding your other comment: Using `bash -c ...` is also not platform independent.

.pre-commit-config.yaml
DoPri: NB: `cp` via `bash -c` isn't platform-independent either, it breaks on Windows. And the hook only exists to keep the committed duplicate in sync, so once gradle reads `api/openapi.yaml` directly (see the other thread) both the hook and the second file go away.

.pre-commit-config.yaml
DoPri: My concern: Also openapi-typescript is now pinned twice (`^7.13.0` in package.json vs `7.13.0` here) with two different call styles, so they can drift. This should be de-duplicated. If you want to move it out of `package.json`, I'd also be fine with that.

services/spring/openapi.yaml
ladu-tu: I already played around with changing the context for the spring services to the project root in the `docker-compose.yml`. That is also suboptimal, as it requires to modify e.g. the `.dockerignore` to include all (also possibly newly added) directories like `services/client`, `docs` etc., otherwise the build context gets bloated. This is something that could easily be forgotten when performing completely unrelated changes. So this approach would also be suboptimal and, at least in my opinion, even worse.

I haven't heard about Docker named build contexts, I'll check it out.

.pre-commit-config.yaml
ladu-tu: I moved it out of `package.json`.

services/client/src/api/schema.d.ts
```diff
@@ -796,7 +1462,7 @@ export interface operations {
             };
         };
     };
-    getDocument: {
+    getSummary: {
```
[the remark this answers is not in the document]
ladu-tu: Fixed.

services/spring/knowledgebase-service/build.gradle
```diff
@@ -11,6 +12,37 @@ openApi {
...
+		useTags             : "true",
+		hideGenerationTimestamp: "true",
+		annotationLibrary   : "none",
+		documentationProvider: "none",
+	]
+}
+
+sourceSets.main.java.srcDir("${genaiGeneratedDir}/src/main/java")
+tasks.named("compileJava").configure { dependsOn(tasks.named("openApiGenerate")) }
+
```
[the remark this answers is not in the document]
ladu-tu: Fixed.

services/spring/qa-service/build.gradle
```diff
@@ -10,3 +11,34 @@ openApi {
...
+		performBeanValidation: "false",
+		useTags             : "true",
+		hideGenerationTimestamp: "true",
+		annotationLibrary   : "none",
+		documentationProvider: "none",
+	]
+}
+
+sourceSets.main.java.srcDir("${genaiGeneratedDir}/src/main/java")
+tasks.named("compileJava").configure { dependsOn(tasks.named("openApiGenerate")) }
```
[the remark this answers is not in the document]
ladu-tu: Fixed.

services/spring/knowledgebase-service/build.gradle
ladu-tu: Fixed.

DoPri (changes requested):
Please re-request a review once the CI is green.

DoPri (approved):
Thank you! One little thing left: Add a small unit test for the `extract()` mapping (unknown/`null` type filtering + `toDouble`), then this is good to merge.

PR #292, opened 2026-07-11
Task: Improve exception handling and validation for Spring microservices

ladu-tu:
## What does this PR do?
This PR closes #287. Although strictly not in the scope of that issue, this includes also a refactor of the exception handling system by moving common exception handlers to a `common` module that is then included and extended by the respective services. Not only the `MethodArgumentNotValidException` and `HttpMessageNotReadableException` have been added, but also additional handlers for other exceptions that may be thrown. As a final point, I also added further logging and improved the `ErrorResponse` schema.

## Type of change

- [ ] Bug fix
- [X] New feature
- [X] Refactor / cleanup
- [ ] Docs
- [X] CI / infra

## Definition of Done

- [x] CI is green
- [X] If the API changed: `api/openapi.yaml` is updated and lint passes
- [X] If a service was added or restructured: docker-compose still comes up cleanly (`docker compose up`)
- [X] Tests added or updated for the changed behaviour
- [X] Docs updated where it matters (README, ADRs, comments)

## Summary by CodeRabbit

* **New Features**
  * Standardized JSON error payloads with field-level validation details across Spring services.
  * Added shared exception handling for common HTTP/auth failures and improved request sanitization.
  * Enforced AskRequest validation: question is required, non-blank, and limited to 1–1500 characters.

* **Bug Fixes**
  * Aligned OpenAPI error schemas with actual service responses, improving consistency for malformed/unsupported requests.

* **Documentation**
  * Updated OpenAPI version to 4.0.0 and refreshed the error contract definitions.

* **Tests**
  * Added/updated tests covering validation, authentication/authorization, and error handling.


api/openapi.yaml
```diff
@@ -1216,6 +1222,10 @@ components:
       properties:
         question:
           type: string
+          maxLength: 1500
+          minLength: 0
```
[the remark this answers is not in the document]
ladu-tu: Fixed.

services/spring/common/src/main/java/com/alexandria/common/web/BaseExceptionHandler.java
```diff
@@ -0,0 +1,109 @@
...
+        log.warn("IllegalArgumentException at {} {}: {}", request.getMethod(), request.getRequestURI(), e.getMessage());
+        return ResponseEntity.badRequest().body(new ErrorResponse("BAD_REQUEST", e.getMessage()));
+    }
+
+    @ExceptionHandler(MethodArgumentNotValidException.class)
+    public ResponseEntity<ErrorResponse> handleValidation(MethodArgumentNotValidException e) {
+        List<FieldError> fieldErrors = e.getBindingResult().getFieldErrors().stream().map(err -> new FieldError(err.getField(), err.getDefaultMessage())).toList();
+        log.debug("Validation failed: {}", fieldErrors);
+        return ResponseEntity.badRequest().body(new ErrorResponse("VALIDATION_ERROR", "Request validation failed", fieldErrors));
+    }
```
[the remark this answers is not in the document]
ladu-tu: Fixed.

services/spring/common/src/main/java/com/alexandria/common/web/BaseExceptionHandler.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @ExceptionHandler(IllegalArgumentException.class)
+    public ResponseEntity<ErrorResponse> handleBadRequest(IllegalArgumentException e, HttpServletRequest request) {
+        log.warn("IllegalArgumentException at {} {}: {}", request.getMethod(), request.getRequestURI(), e.getMessage());
+        return ResponseEntity.badRequest().body(new ErrorResponse("BAD_REQUEST", e.getMessage()));
+    }
+
+    @ExceptionHandler(MethodArgumentNotValidException.class)
+    public ResponseEntity<ErrorResponse> handleValidation(MethodArgumentNotValidException e) {
+        List<FieldError> fieldErrors = e.getBindingResult().getFieldErrors().stream().map(err -> new FieldError(err.getField(), err.getDefaultMessage())).toList();
```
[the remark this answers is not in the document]
ladu-tu: Fixed.

api/openapi.yaml
DoPri, line 1: No version bump?

services/spring/common/src/main/java/com/alexandria/common/web/BaseExceptionHandler.java
```diff
@@ -0,0 +1,122 @@
...
+    public ResponseEntity<ErrorResponse> handleAccessDenied(AccessDeniedException e, HttpServletRequest request) {
+        log.warn("Access denied at {} {} (user={}): {}", sanitize(request.getMethod()), sanitize(request.getRequestURI()), sanitize(request.getRemoteUser()), sanitize(e.getMessage()));
+        return ResponseEntity.status(HttpStatus.FORBIDDEN).body(new ErrorResponse("FORBIDDEN", "Access denied"));
+    }
+
+    @ExceptionHandler(AuthenticationException.class)
+    public ResponseEntity<ErrorResponse> handleAuthentication(AuthenticationException e, HttpServletRequest request) {
+        log.warn("Authentication failed at {} {}: {}", sanitize(request.getMethod()), sanitize(request.getRequestURI()), sanitize(e.getMessage()));
+        return ResponseEntity.status(HttpStatus.UNAUTHORIZED).body(new ErrorResponse("UNAUTHORIZED", "Authentication required"));
+    }
```
DoPri, line 115: `handleAccessDenied` / `handleAuthentication` are effectively unreachable in the current setup. All three services use `oauth2ResourceServer`. For a resource server, `AuthenticationException`s and `AccessDeniedException`s are thrown in the security filter chain and handled by `BearerTokenAuthenticationEntryPoint` / `BearerTokenAccessDeniedHandler` *before* the DispatcherServlet, so `@RestControllerAdvice` never sees them. No `@PreAuthorize`/`@EnableMethodSecurity` exists in the codebase, so these two handlers won't fire in practice, and 401/403 responses will not use the unified `ErrorResponse` schema.

services/spring/common/src/main/java/com/alexandria/common/web/ErrorResponse.java
```diff
@@ -0,0 +1,13 @@
+package com.alexandria.common.web;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import java.util.List;
+
+/** Unified error payload returned by every Alexandria Spring service. */
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
DoPri, line 8: Add `@JsonInclude(NON_NULL)` to `FieldError` as well for consistency

services/spring/Dockerfile
```diff
@@ -18,6 +18,7 @@ COPY knowledgebase-service/build.gradle knowledgebase-service/build.gradle
 COPY qa-service/build.gradle qa-service/build.gradle
 # hadolint ignore=DL3022
 COPY --from=api openapi.yaml ${SERVICE}/build/openapi-input/openapi.yaml
+COPY common/ common/
```
DoPri: Sits before `RUN ./gradlew :${SERVICE}:dependencies` and copies the whole module (incl. `src/` and `src/test/`). This breaks the layer-caching pattern used just above (L16-18 copy only each service's `build.gradle` first, then `src` at L25): any change to `common` source now invalidates the dependency-resolution layer and everything after it. Consider `COPY common/build.gradle common/build.gradle` before `dependencies`, then `COPY common/src common/src` before `bootJar`.

services/spring/user-service/build.gradle
```diff
@@ -11,3 +11,7 @@ openApi {
 		args = ["--spring.profiles.active=openapi"]
 	}
 }
+
+dependencies {
+	implementation project(':common')
+}
```
DoPri, line 17: `common` inherits the full service dependency set from the root `subprojects {}` block even though it's a plain web-utility JAR. This makes the module is much heavier than it needs to be. Optional cleanup: scope the heavy deps out of `common`, or declare only what the base handler needs (web, validation, security-core, slf4j, jackson).

DoPri (changes requested):

services/spring/user-service/build.gradle
ladu-tu: Yes, that was a deliberate choice by me. The inherited dependencies don't actually cause any problems (except for the larger size), but I think restructuring that takes more work than is warranted here. Or is there an approach I don't know about?

services/spring/common/src/main/java/com/alexandria/common/web/BaseExceptionHandler.java
ladu-tu: I know that these handlers are not reachable in our current setup. I added them as "fallbacks". The moment we add `@EnableMethodSecurity/@PreAuthorize` (or any controller-level auth check), these will start producing the unified ErrorResponse. Until then, they don't really incur any penalty/overhead. I would suggest leaving this as-is.

api/openapi.yaml
ladu-tu: Bumped.

services/spring/user-service/build.gradle
ladu-tu: To finalize: I'd rather keep the build config simple and uniform. So I would leav this as-is. If the module footprint ever becomes a real concern, we can revisit this.

services/spring/Dockerfile
ladu-tu: Valid.

services/spring/common/src/main/java/com/alexandria/common/web/ErrorResponse.java
ladu-tu: Added.

api/openapi.yaml
DoPri: Doesn't look bumped to me.

services/spring/common/src/main/java/com/alexandria/common/web/BaseExceptionHandler.java
DoPri: Still, the problem remains: 401/403 responses will not use the unified ErrorResponse schema.

DoPri (changes requested):

DoPri (approved):
Thank you.

PR #299, opened 2026-07-13
CI: Add per-service test coverage reports and badges

DoPri:
## What does this PR do?

Adds per-service test coverage reporting and wires up coverage badges. Refs #228.

- Spring: JaCoCo on every module. Tests finalize with `jacocoTestReport` and emit XML, so the `./gradlew build` CI already runs also produces the report. The OpenAPI-generated GenAI client is excluded so it doesn't drag the numbers down.
- GenAI: added `pytest-cov` and `--cov=app --cov-report=xml` to the pytest defaults, so every `pytest` run (local, CI, and the `genai-test` container) drops a `coverage.xml`.
- CI: an upload step in `build-spring` (one flag per module via the matrix) and `build-genai` (flag `genai`), pinned to `codecov-action` v7.
- `codecov.yml`: per-service flags with carryforward (a PR touching one service doesn't blank the others' badges), path scoping, coverage kept informational (reports, doesn't gate merges, same stance as trivy/kics).
- README: coverage table with badges for user-service, knowledgebase-service, qa-service, genai.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Docs
- [x] CI / infra

## Definition of Done

- [ ] CI is green
- [x] If a service was added or restructured: docker-compose still comes up cleanly (`docker compose up`) (no structural change; genai-test still `uv sync`s dev deps incl. pytest-cov)
- [x] Tests added or updated for the changed behaviour (coverage tooling per service; verified JaCoCo XML for common + qa-service and genai coverage.xml locally)

## Notes for the reviewer

Draft on purpose. There's one thing that isn't in our hands: for the badges to actually populate, the repo has to be activated on codecov.io and a `CODECOV_TOKEN` secret added, both of which need org/repo admin. Uploads use `fail_ci_if_error: false`, so CI stays green either way; the badges just show "unknown" until that's sorted.

I looked into whether tokenless would let us skip the admin step since the repo is public. It doesn't: genuine tokenless only covers fork PRs into an upstream, and the public-repo "token not required" opt-out is an org-level setting on codecov.io that only an admin can flip (and it still needs the Codecov app installed on the org). See codecov-action#1845 for the same situation. If we can't get admin to do it, the fallback is self-hosting the badges from CI on a `badges` branch using the built-in token, no external service. Happy to switch to that.

Client badge is left out for now: the client only has Playwright e2e, no unit suite to measure, so there's nothing to turn into an lcov report yet. Noted in the README; can add it once the client has unit tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -317,6 +317,18 @@ jobs:
         working-directory: services/spring
         run: ./gradlew --no-daemon --stacktrace :${{ matrix.service.gradle_module }}:build
 
+      - name: Upload Coverage to Codecov
+        # JaCoCo writes the XML as a finalizer of the test task, so `build` above
+        # already produced it. One flag per module -> one badge per service.
+        if: always() && env.RUN_JOB == 'true'
+        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
+        with:
+          files: services/spring/${{ matrix.service.gradle_module }}/build/reports/jacoco/test/jacocoTestReport.xml
+          flags: ${{ matrix.service.name }}
+          disable_search: true
+          fail_ci_if_error: false
+          token: ${{ secrets.CODECOV_TOKEN }}
+
       - name: Upload Test Reports
         if: always() && env.RUN_JOB == 'true'
         uses: actions/upload-artifact@v7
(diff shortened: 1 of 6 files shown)


DoPri:
Implemented workaround in #307

PR #302, opened 2026-07-13
Feat: Add autoscaling, NetworkPolicies, and Prometheus docs to Helm char

ladu-tu:
## What does this PR do?

This PR addresses the main helm issues still open. Closes #282, closes #283 and closes #284.
- Regarding #282: Adds autoscaling to all stateless services: spring microservices, genai, client. Gated by `autoscaling.enabled`, disabled by default.
- Regarding #283: Added default-deny ingress, default-allow egress as NetworkPolicy. Gated by `networkPolicy.enabled`, disabled by default.
- Reagrding #284: Changed scrape configs to be more closely aligned between `docker-compose` setup and helm deployment. Complete sync not possible because of different integration of `traefik` and `s3-storage`. Added a small note to the README on why we keep our static scrape config, TL;DR: Missing permissions.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] CI / infra

## Definition of Done

- [x] CI is green

## Summary by CodeRabbit

* **New Features**
  * Added optional per-service CPU autoscaling with configurable replica limits.
  * Added optional network policies with default-deny ingress and explicit service access rules.
  * Added Kubernetes deployment guidance for autoscaling, networking, and monitoring configuration.

* **Monitoring**
  * Replaced the GenAI latency alert with alerts for slow model calls and elevated model-call failure rates.

### DIFF
--- a/infra/k8s/README.md
+++ b/infra/k8s/README.md
@@ -107,17 +107,17 @@ When a service's autoscaling is enabled the HPA owns the replica count, so the s
 kubectl -n alexandria get hpa
 ```
 
-### Rationale on Static Scrape Config
+## Rationale on Static Scrape Config
 
 We run our own Prometheus and Grafana as plain deployments and feed Prometheus a static scrape config from a ConfigMap (`templates/prometheus-config.yaml`), rather than using the Prometheus operator with ServiceMonitor/PodMonitor and PrometheusRule CRDs.
 
-The operator isn't installed on the stud cluster and the CRDs aren't available, thus ServiceMonitors would need us to bring up and manage the whole operator stack. For a fixed, small set of services that only changes when we add a service, a hand-written scrape config is simpler, needs no extra cluster components, and keeps the k8s setup close to the docker-compose one.
+The `monitoring.coreos.com` CRDs exist on the stud cluster, but that operator and its Prometheus are cluster-owned. We don't control its scrape scope, alerting, or Grafana and shouldn't rely on shared infrastructure for our project's monitoring. Running our own Prometheus + Grafana with a self-contained static scrape config keeps monitoring fully owned by the chart, needs no coordination with the cluster operator, and mirrors the docker-compose setup. For a fixed, small set of services that only changes when we add a service, a hand-written scrape config is also simpler than wiring ServiceMonitors into our own operator install.
 
 The scrape targets differ slightly from the compose setup on purpose: compose runs an all-in-one SeaweedFS container and Traefik, while the chart splits SeaweedFS into `seaweedfs-s3` and `seaweedfs-volume` (different metrics port) and fronts traffic with the ingress instead of Traefik. The alert rules are kept in sync with `infra/prometheus/alert.rules.yml`.
 
 ## NetworkPolicies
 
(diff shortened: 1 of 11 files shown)


ladu-tu:
@coderabbitai review

infra/k8s/alexandria/templates/networkpolicies.yaml
```diff
@@ -0,0 +1,263 @@
...
+    matchExpressions:
+      - key: app
+        operator: In
+        values: ["client", "user-service", "knowledgebase-service", "qa-service", "genai"]
+  ingress:
+    - from:
+        - namespaceSelector:
+            {{- toYaml $ic.namespaceSelector | nindent 12 }}
+          podSelector:
+            {{- toYaml $ic.podSelector | nindent 12 }}
```
DoPri: With the `values.yaml` defaults (`ingressController.namespaceSelector: {}`, `podSelector: {}`) this `from` peer matches all pods in all namespaces. So `--set networkPolicy.enabled=true` with nothing else opens client + all 4 services to the whole cluster, which quietly undoes the default-deny. And it won't be noticed because real ingress traffic keeps working. You should require real selectors (fail the render when empty).

infra/k8s/README.md
```diff
@@ -88,6 +88,50 @@ kubectl -n alexandria get svc
...
+kubectl -n alexandria get hpa
+```
+
+### Rationale on Static Scrape Config
+
+We run our own Prometheus and Grafana as plain deployments and feed Prometheus a static scrape config from a ConfigMap (`templates/prometheus-config.yaml`), rather than using the Prometheus operator with ServiceMonitor/PodMonitor and PrometheusRule CRDs.
+
+The operator isn't installed on the stud cluster and the CRDs aren't available, thus ServiceMonitors would need us to bring up and manage the whole operator stack. For a fixed, small set of services that only changes when we add a service, a hand-written scrape config is simpler, needs no extra cluster components, and keeps the k8s setup close to the docker-compose one.
+
+The scrape targets differ slightly from the compose setup on purpose: compose runs an all-in-one SeaweedFS container and Traefik, while the chart splits SeaweedFS into `seaweedfs-s3` and `seaweedfs-volume` (different metrics port) and fronts traffic with the ingress instead of Traefik. The alert rules are kept in sync with `infra/prometheus/alert.rules.yml`.
```
DoPri: This does not really belong in the Autoscaling section. Rather in the Monitoring section, I'd say.

DoPri (changes requested):

infra/k8s/README.md
ladu-tu: That was just the wrong heading size.

infra/k8s/alexandria/templates/networkpolicies.yaml
ladu-tu: Good catch, fixed. The render now fails fast when the ingress-controller selectors are empty.

While wiring this up I hit a second problem: on the stud cluster we have no permission to discover the ingress controller, so we could only guess the installed controller on the cluster, which blocks all external traffic if we guess wrongly (tried it out, got a Gateway Timeout).

So I added `networkPolicy.ingressController.allowFromAnywhere` with default true, so only client, geani docs, spring services and keycloak auth accept connections from any source. The other pods stay locked down.

DoPri (approved):
Thank you. Good work!

PR #318, opened 2026-07-14
GenAI: add opt-in OpenTelemetry tracing

DoPri:
## What does this PR do?

Adds OpenTelemetry tracing to the GenAI service so a request shows up as a span tree instead of just a latency number. The FastAPI request span is the parent, and every model/embedding call nests underneath it, so on a slow `/genai/ask` you can finally see whether the time went into the query embedding or the chat call.

Tracing is opt-in and off in the default stack. It only exports when `OTEL_EXPORTER_OTLP_ENDPOINT` is set; otherwise the spans run on OpenTelemetry's no-op tracer and cost effectively nothing. A new `docker-compose.tracing.yml` overlay wires it to a local Jaeger:

```bash
docker compose -f docker-compose.yml -f docker-compose.tracing.yml up
```

Jaeger UI is then at http://localhost/jaeger/ (service `alexandria-genai`).

This ticks the "LangChain tracing to OpenTelemetry" box from the observability checklist in #219 (not closing the whole tracking issue).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Docs
- [x] CI / infra

## Definition of Done

- [x] CI is green

## Notes for the reviewer

A few decisions worth a look:

- Spans live in the one place every provider call already funnels through, `run_model_call` in `app/model_calls.py`, so there's a single wrap point rather than instrumentation scattered across summarize/extract/tag/qa/search. The model spans carry the OTel GenAI attributes (`gen_ai.operation.name`, `gen_ai.system`, `gen_ai.request.model`, `gen_ai.response.model`) plus a `genai.outcome` that mirrors the existing metric status label.
- I did not pull in a LangChain callback instrumentation library (openinference/openllmetry). They don't support langchain 1.x yet, and the chokepoint wrapper gives us what we need without the version risk. If we want the full LangChain run tree later, that can be a follow-up once the libs catch up.
- Jaeger uses OTLP over HTTP (port 4318), not gRPC, to avoid dragging `grpcio` into the image (no cp314 wheel story I wanted to bet on).

To try it: bring up the overlay, hit summarize/ask/search a few times, then look at the `alexandria-genai` service in Jaeger. `/genai/health` and `/genai/metrics` are excluded so the probe and scrape traffic doesn't drown out the real requests.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in OpenTelemetry tracing for GenAI requests and model calls.
  * Added configurable OTLP endpoint and service name settings.
  * Added a Docker Compose tracing setup with local Jaeger visualization.
  * Excluded health and metrics endpoints from tracing.

* **Documentation**
  * Added setup instructions, configuration guidance, trace details, and Jaeger access information.

* **Tests**
  * Added coverage for tracing activation, span attributes, successful calls, errors, and timeouts.

### DIFF
--- a/docker-compose.tracing.yml
+++ b/docker-compose.tracing.yml
@@ -1,0 +1,40 @@
+# Opt-in tracing overlay. Adds a local Jaeger and points the GenAI service's
+# OTLP exporter at it. Not part of the default stack, bring it up explicitly:
+#
+#   docker compose -f docker-compose.yml -f docker-compose.tracing.yml up
+#
+# Then open http://localhost/jaeger/, pick the "alexandria-genai" service, and
+# hit an AI endpoint (summarize, ask, search) to generate traces. Each request
+# shows up with its model/embedding calls as child spans.
+
+services:
+  genai:
+    environment:
+      OTEL_EXPORTER_OTLP_ENDPOINT: ${OTEL_EXPORTER_OTLP_ENDPOINT:-http://jaeger:4318}
+      OTEL_SERVICE_NAME: ${OTEL_SERVICE_NAME:-alexandria-genai}
+    depends_on:
+      jaeger:
+        condition: service_started
+
+  jaeger:
+    image: jaegertracing/all-in-one:1.62.0
+    container_name: jaeger
+    # kics-scan ignore-block
+    environment:
+      COLLECTOR_OTLP_ENABLED: "true"
+      # Serve the UI under /jaeger so Traefik can route it like the other tools.
+      QUERY_BASE_PATH: /jaeger
+    restart: unless-stopped
+    networks:
+      - alexandria
+    deploy:
+      resources:
+        limits:
+          cpus: "0.50"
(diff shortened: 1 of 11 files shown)


docker-compose.tracing.yml
ladu-tu, line 1: Do we really want a specific `docker-compose` for this tracing alone? Until now, we've kept everything in one file.

ladu-tu (approved):
Nice work! Just one question.

docker-compose.tracing.yml
DoPri: At least from what I tried, this "override" pattern with the compose files is the "cleanest" solution. Adding it to the main compose file would essentially lead to us having two separate GenAI configs that we would have to maintain simultaneously. (compared to e.g., the e2e tests, which we only need to run during development, tracing should also be possible in production).
If you have a idea, how I could implement that in one compose file, I'm definitely open to that.

PR #270, opened 2026-07-10
Feat: Integrate GenAI tagging and semantic search functionality on Spring side

ladu-tu:
## What does this PR do?

This PR adresses and closes #247, closes #245. It consists of two parts, that due to their similarity, I added to one PR.

Tagging: The `knowledgebase-service` now calls the tagging endpoint on the GenAI side on document upload/creation. We therefore now also have auto-generated tags. I also added an endpoint to regenerate the tags (redoes the GenAI service request), to mirror the other similar endpoints (summarization and entity extraction). This regeneration endpoint keeps user-provided tags though, as it seemed reasonable to not remove a user's manually assigned tags.

Semantic Search: The `knowledgease-service` already supported a search endpoint without GenAI interaction, namely keyword-based. I expanded that endpoint to not only search in document names, but also the extracted `rawTextContent` (closes #289). I moved this endpoint to `/knowledgebase/search/text`. The second search endpoint is at `/knowledgebase/search/semantic`, which uses the GenAI service for semantic search, as requested in #247.

Docs and tests updated and added accordingly.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor / cleanup
- [X] Docs
- [ ] CI / infra

## Definition of Done

- [x] CI is green
- [X] If the API changed: `api/openapi.yaml` is updated and lint passes
- [X] If a service was added or restructured: docker-compose still comes up cleanly (`docker compose up`)
- [X] Tests added or updated for the changed behaviour
- [X] Docs updated where it matters (README, ADRs, comments)

## Notes for the reviewer

Needs to wait until #261 is merged.

## Summary by CodeRabbit

* **New Features**
  * Added a semantic search endpoint returning relevance scores and matched text snippets, with configurable `limit` and explicit indication when semantic fallback to keyword search is used.
  * Split keyword search into a dedicated text endpoint matching across filenames and document content, returning lightweight document details.
  * Added an endpoint to reprocess document tags (replacing only auto-generated tags while keeping user-added tags).
* **Documentation**
  * Updated OpenAPI/versioned API documentation and response shapes for the new search and reprocessing endpoints.

PR #321, opened 2026-07-14
Feat: Add search UI with semantic and keyword modes

BjarneHa:
## What does this PR do?

Adds a Search tab to the client, closes #276, closes #248.

The tab lives at `/search` and lets users search their documents in two modes -- semantic (default) and keyword -- via the endpoints introduced in #270. Semantic results show a relevance score and the matching text snippet from the document. Keyword results show file type and size as secondary metadata. The current query and mode are stored in the URL as `?q=` and `?mode=`, so searches are shareable and survive back/forward navigation.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Docs
- [ ] CI / infra

## Definition of Done

- [x] CI is green

## Notes for the reviewer

The error tests for search use a 15s timeout instead of the usual 5s because the search endpoints use `useQuery` (not `useMutation`), which retries up to 3 times before surfacing the error -- by design, so a transient 401 triggers re-auth and the retry succeeds. With a permanently-stubbed failure the full retry backoff is ~7s.

## Summary by CodeRabbit

* **New Features**
  * Added a Search page with semantic and keyword search modes.
  * Added relevance scores, snippets, result links, empty states, and fallback notices when semantic search is unavailable.
  * Added Search navigation and authenticated routing.
* **Bug Fixes**
  * Improved document size formatting and handling of missing values.
* **Documentation**
  * Documented search modes, result behavior, and fallback handling.
* **Tests**
  * Added coverage for search navigation, queries, results, errors, authentication, and fallback behavior.

### DIFF
--- a/services/client/src/components/SearchPanel.tsx
+++ b/services/client/src/components/SearchPanel.tsx
@@ -1,0 +1,304 @@
+import React, { useEffect, useState } from "react";
+import { Link, useSearchParams } from "react-router";
+import $api from "../api/client";
+import type { components } from "../api/schema";
+
+type SemanticSearchResultDto = components["schemas"]["SemanticSearchResultDto"];
+type DocumentRefDto = components["schemas"]["DocumentRefDto"];
+
+/**
+ * Maps a fetch error to a user-facing message, following the same pattern used
+ * in QAPanel and DocumentDetail.
+ */
+function errorMessage(
+  err: unknown,
+  messages: { notAuthenticated: string; forbidden: string; fallback: string },
+): string {
+  const code = err instanceof Error ? err.message : String(err);
+  if (code === "NOT_AUTHENTICATED") return messages.notAuthenticated;
+  if (code === "FORBIDDEN") return messages.forbidden;
+  return messages.fallback;
+}
+
+function formatFileSize(bytes: number | undefined): string {
+  if (bytes == null) return "";
+  if (bytes < 1024) return `${bytes} B`;
+  if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} KB`;
+  return `${(bytes / (1024 * 1024)).toFixed(1)} MB`;
+}
+
+function formatFileType(mimeType: string | undefined): string {
+  if (!mimeType) return "";
+  const known: Record<string, string> = {
+    "application/pdf": "PDF",
+    "text/plain": "Text",
+    "text/markdown": "Markdown",
+    "application/msword": "Word",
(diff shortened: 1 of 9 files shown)


services/client/src/components/SearchPanel.tsx
```diff
@@ -0,0 +1,304 @@
...
+  const {
+    data: textData,
+    isLoading: textLoading,
+    error: textError,
+  } = $api.useQuery(
+    "get",
+    "/api/v1/knowledgebase/search/text",
+    { params: { query: { q: submittedQuery } } },
+    { enabled: mode === "text" && submittedQuery.length > 0 },
+  );
```
DoPri, line 219: Search uses `useQuery`, which retries 3x on *every* error, not just 401. On a 500 the user sits on "Searching..." with the input disabled for the full backoff (~7s) before the error shows. Only the NOT_AUTHENTICATED case actually benefits from a retry. Consider a `retry` predicate that only retries on re-auth, e.g. `retry: (n, e) => e.message === "NOT_AUTHENTICATED" && n < 3`. That also fixes the locked-input UX and lets the error tests drop the 15s timeout.

services/client/src/components/SearchPanel.tsx
```diff
@@ -0,0 +1,304 @@
...
+  if (code === "FORBIDDEN") return messages.forbidden;
+  return messages.fallback;
+}
+
+function formatFileSize(bytes: number | undefined): string {
+  if (bytes == null) return "";
+  if (bytes < 1024) return `${bytes} B`;
+  if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} KB`;
+  return `${(bytes / (1024 * 1024)).toFixed(1)} MB`;
+}
```
DoPri: `formatFileSize` is a copy of `formatBytes` in DocumentDetail.tsx:14 (only the null return differs). Move one into `utils/format.ts` and share it.

DoPri (changes requested):

services/client/src/components/SearchPanel.tsx
BjarneHa: Wouldn't some 500s also benefit from a retry? Assume the genai service times out on the response from Logos. This would cause a 500. The next request could succeed if Logos replies faster.

DoPri: Yes, that's also acceptable. Just wanted to mention it. Not a show stopper.

DoPri (approved):

PR #324, opened 2026-07-15
Fix: Calibrate GenAI search scores and stop reindex dropping chunks

DoPri:
## What does this PR do?

Fixes two things that made semantic search and Ask look broken on real documents.

**Search relevance scores were unreadable.** We were showing raw cosine similarity as a percentage. Qwen3-Embedding cosine sits in a narrow band (unrelated text still scores ~0.15, a strong keyword hit ~0.35), so a genuine match read ~30% right next to ~20% noise and looked like nothing matched. I checked the embeddings first: the Logos gateway reproduces the Qwen3 model card reference pairs to three decimals, and the Weaviate collection really is on cosine distance, so the embeddings are fine, only the display was wrong. `_score` now calibrates cosine `[FLOOR..CEILING] -> [0..1]` (monotonic, so ranking is unchanged), with the anchors env-configurable via `SEARCH_SCORE_FLOOR`/`SEARCH_SCORE_CEILING` (defaults 0.15/0.45, tuned for the short keyword queries search actually gets). Query embedding also now uses the Qwen3 retrieval instruction, which is the documented usage.

On the test doc that meant: verbatim matches 45-62%, a weaker match 31%, gibberish 9%, unrelated 0%.

**Ask sometimes answered from an incomplete document.** Asking "Wann Warnstreik?" got "kein Datum" even though the document clearly states the date. Root cause was indexing, not the LLM: `index_chunks` reused deterministic Weaviate ids, so re-indexing (delete + reinsert the same id) could reinsert a chunk over the tombstone the delete just wrote, and until async tombstone cleanup runs that chunk is missing from the vector index while still existing as an object. The chunk holding the date was invisible to retrieval, so the model answered honestly from what it got. Chunks now insert under fresh ids (the delete already prevents duplicates), so re-index never lands on a tombstone.

Because documents indexed by the old scheme can stay half-indexed until rebuilt, this also adds `app/reindex.py` (`docker compose exec genai python -m app.reindex`) to rebuild every currently-indexed document once from object storage.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Docs
- [ ] CI / infra

## Definition of Done

- [x] CI is green

## Notes for the reviewer

- The score calibration constants are empirically derived and commented in `search.py`; the anchors are env vars (wired into compose, `.env.example`, the k8s configmap and values) so they can be tuned per embedding model without a code change.
- The fresh-id change in `vectorstore.py` has a comment explaining why deterministic ids are deliberately avoided, so nobody "restores" them for idempotency and silently reintroduces the missing-chunk bug.
- After merge/deploy, run `docker compose exec genai python -m app.reindex` once to heal documents indexed under the old id scheme.

## Summary by CodeRabbit

* **New Features**
  * Calibrated `/genai/search` relevance scores using configurable `SEARCH_SCORE_FLOOR` and `SEARCH_SCORE_CEILING`.
  * Model-specific query/document formatting for embeddings (improves semantic matching).
  * Added a one-off reindex utility to heal and rebuild half-indexed documents.
* **Documentation**
  * Updated GenAI docs to explain score calibration, new environment variables, and indexing/recovery behavior.
* **Bug Fixes**
  * Improved reindex replacement behavior to ensure chunk reinsertion works reliably.
* **Tests**
  * Expanded coverage for score calibration, embeddings formatting, vectorstore key listing, and reindex behavior.


DoPri:
@coderabbitai review

ladu-tu (changes requested):
LGTM except for some minor questions:
1. Should we show some warning / fail if `ceiling <= floor`?
2. Your PR description states that "verbatim matches 45-62%". Shouldn't that be a bit higher for _verbatim_ matches? Like 80%-100%? Consider adapting the ceiling.

DoPri:
> LGTM except for some minor questions:
> 
>     1. Should we show some warning / fail if `ceiling <= floor`?
> 
>     2. Your PR description states that "verbatim matches 45-62%". Shouldn't that be a bit higher for _verbatim_ matches? Like 80%-100%? Consider adapting the ceiling.

Fair questions:
1. Will fix.
2. I actually also thought about that. Here's my reasoning: These are semantic cosine scores, not lexical, so "verbatim" doesn't usually map to ~100%. With Qwen a single word against a full multi-topic chunk caps around 0.3-0.4 cosine no matter how exact it looks (see model sheet for Qwen). 0.6+ is only full-sentence matches since this is what semantic search is really supposed to be. To make "Warnstreik" read 90% I'd have to drop the ceiling to 0.3, which then saturates the genuinely strong matches at 100% too and flattens the gradient. Exact-term = 100% is really the Keyword (BM25) search's job. Having said that, the ceiling is just an env var (SEARCH_SCORE_CEILING), so users can tune it to their liking.

ladu-tu (approved):
Okay, thanks for the clarification!

PR #325, opened 2026-07-15
Infra: switch helm secrets to existingSecret pattern with auto-generation

BjarneHa:
## What does this PR do?

Replaces the old `values-secrets.yaml` / `--set` approach for secret management with the `existingSecret` pattern. Closes #306.

On a bare `helm install`, all credentials except the LLM API key are now randomly generated and persisted across upgrades via `lookup`. No extra files, no `--set` flags for passwords. The only required input is `--set genai.llmApiKey=...`.

For CI, secrets are pre-created via `kubectl` before Helm runs so they never appear in release metadata. Helm is told about them with `--set existingSecret=...` and `--set genai.existingSecret=...`, which suppresses the chart's own generation.

The chart produces four secrets from a single template file (`secrets.yaml`), sharing resolved variables so values are always consistent:

- `alexandria-secrets` -- primary, all auto-generated credentials
- `alexandria-postgres-auth` -- derived, copies `db-password` and hardcodes `db-username` to "postgres" (prevents override)
- `alexandria-seaweedfs-s3-config` -- derived, builds the SeaweedFS identity JSON from S3 keys
- `alexandria-grafana` -- derived, copies the Grafana admin password into the monitoring namespace (cross-ns pod constraint)

`alexandria-genai-secrets` is separate since the LLM key is an external credential that can't be generated.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Docs
- [x] CI / infra

## Definition of Done

- [ ] CI is green

## Notes for the reviewer

The `values-secrets.yaml.example` file is deleted since it's fully replaced by the README docs and the auto-generation logic.

## Summary by CodeRabbit

* **New Features**
  * Added centralized Kubernetes Secret management for database, storage, Grafana, Keycloak, internal service, and GenAI credentials.
  * Credentials can now be supplied through pre-created Secrets or generated and preserved automatically across upgrades.
  * Helm deployments now support stable, reusable Secret references.

* **Documentation**
  * Updated installation, upgrade, uninstall, secret-management, and troubleshooting guidance for the new deployment workflow.


infra/k8s/alexandria/templates/secrets.yaml
```diff
@@ -0,0 +1,87 @@
...
+     existingSecret is set, $existing is the kubectl-created secret, and the primary
+     document below is suppressed -- but the derived documents still get consistent
+     values from the same variables. */}}
+{{- $dbPassword := .Values.spring.database.password | default ($existing | dig "data" "db-password" "" | b64dec) | default (randAlphaNum 24) }}
+{{- $s3AccessKey := .Values.s3.accessKey | default ($existing | dig "data" "s3-access-key" "" | b64dec) | default (randAlphaNum 20) }}
+{{- $s3SecretKey := .Values.s3.secretKey | default ($existing | dig "data" "s3-secret-key" "" | b64dec) | default (randAlphaNum 40) }}
+{{- $internalSecret := .Values.spring.internal.sharedSecret | default ($existing | dig "data" "internal-shared-secret" "" | b64dec) | default (randAlphaNum 32) }}
+{{- $keycloakAdminPassword := .Values.keycloak.keycloak.adminPassword | default ($existing | dig "data" "keycloak-admin-password" "" | b64dec) | default (randAlphaNum 24) }}
+{{- $grafanaAdminPassword := .Values.grafana.adminPassword | default ($existing | dig "data" "grafana-admin-password" "" | b64dec) | default (randAlphaNum 20) }}
```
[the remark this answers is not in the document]
BjarneHa: GitOps is not supported.

infra/k8s/alexandria/templates/secrets.yaml
ladu-tu, line 1: In this file, ech value is `.Values.x | default ($existing | dig ...) | default (randAlphaNum N)`. The persistence relies entirely on `lookup` finding the existing secret. 
However, `lookup` returns empty during `helm template`, `helm install --dry-run`, and anytime the API server can't be reached. If someone runs a plain `helm upgrade` while the cluster is momentarily unreachable, or a tool does a dry-run render and applies it, new passwords are generated and the Postgres PVC is desynced. 

I think it's worth a comment that explicity states that that this template must never be applied from a dry-run render.

infra/k8s/alexandria/templates/secrets.yaml
```diff
@@ -0,0 +1,94 @@
...
+type: Opaque
+stringData:
+  db-password: {{ $dbPassword | quote }}
+  keycloak-admin-password: {{ $keycloakAdminPassword | quote }}
+  grafana-admin-password: {{ $grafanaAdminPassword | quote }}
+  s3-access-key: {{ $s3AccessKey | quote }}
+  s3-secret-key: {{ $s3SecretKey | quote }}
+  internal-shared-secret: {{ $internalSecret | quote }}
+{{- end }}
+{{- if or (not .Values.existingSecret) $existing }}
```
ladu-tu, line 50: If `existingSecret=alexandria-secrets` is set (CI path) but the kubectl secret hasn't been created yet, the derived secrets (`postgres-auth`, `seaweedfs-s3-config`, `grafana`) silently don't render. `keycloak.database.existingSecret: alexandria-postgres-auth` and `postgres-db.existingSecret` then point at a derived secret that doesn't exist, so Keycloak/Postgres fail.

Consider failing loudly when `existingSecret` is set and lookup returns empty, same as you already do for the genai key.

ladu-tu (changes requested):
Have you run a full `helm install` from scratch on a real cluster (not just `helm template`) to confirm the hook secrets land before the postgres/keycloak pods start?

ladu-tu (approved):
Thanks!

PR #326, opened 2026-07-15
Feat: Configurable client API base URL and gateway CORS

DoPri:
## What does this PR do?

Two related bits of config, both around the client talking to the API from a
different origin.

First, the client API base URL is no longer hardcoded to `""`. It now reads
`import.meta.env.VITE_API_URL` and falls back to `""`, so same-origin behaviour
behind Traefik stays exactly as before, but the base URL can be pointed
somewhere else at build time. Closes #278.

Second, since a cross-origin client would immediately hit CORS, I added a CORS
config on both gateways: a Traefik middleware in docker-compose attached to the
three Spring API routers, and the matching nginx-ingress CORS annotations on the
k8s side. The allowed origin is env-driven (`CORS_ALLOWED_ORIGINS` for compose,
`ingress.corsAllowOrigin` in the Helm values) and defaults to `http://localhost`,
so nothing changes for the current same-origin setup. This ticks the CORS box in
the compliance checklist (#219).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] CI / infra

## Definition of Done

- [ ] CI is green
- [ ] Tests added or updated for the changed behaviour (config-only change, no unit test hook; validated with `docker compose config`, `helm template`, and tsc)

## Notes for the reviewer

- Default origin is `http://localhost` on purpose so this is a no-op for the
  current deployment. Override it per environment when the client moves to its
  own origin.
- `VITE_API_URL` is a Vite build-time var, so it gets baked into the image at
  build, not read at container runtime. Documented that in
  `services/client/README.md`.
- Traefik middleware is defined once on the traefik container and referenced as
  `api-cors@docker` from the user/knowledgebase/qa routers.

## Summary by CodeRabbit

* **New Features**
  * Added configurable CORS support for API access from approved web origins.
  * Added build-time configuration for setting a custom API base URL.
  * API requests now use the configured base URL when provided, with same-origin access as the default.

* **Documentation**
  * Documented CORS and API URL configuration, including setup examples and environment variable behavior.

### DIFF
--- a/infra/k8s/alexandria/values.yaml
+++ b/infra/k8s/alexandria/values.yaml
@@ -148,6 +148,10 @@ ingress:
   # nginx client_max_body_size, kept a bit above Spring MAX_UPLOAD_SIZE
   # so oversized uploads get the Spring 413 error response
   maxBodySize: "30m"
+  # Browser origin allowed to call the API cross-origin. Same-origin traffic
+  # through this ingress is unaffected; override when the client is hosted
+  # elsewhere.
+  corsAllowOrigin: "http://localhost"
 
 # NetworkPolicies: default-deny ingress with explicit allow rules.
 # Off by default.
--- a/infra/k8s/alexandria/values.yaml
+++ b/infra/k8s/alexandria/values.yaml
@@ -148,10 +148,10 @@ ingress:
   # nginx client_max_body_size, kept a bit above Spring MAX_UPLOAD_SIZE
   # so oversized uploads get the Spring 413 error response
   maxBodySize: "30m"
-  # Browser origin allowed to call the API cross-origin. Same-origin traffic
-  # through this ingress is unaffected; override when the client is hosted
-  # elsewhere.
-  corsAllowOrigin: "http://localhost"
+  # Browser origin allowed to call the API cross-origin. Set to the ingress
+  # origin so no external origin is granted access; change it when the client is
+  # served from a different origin.
+  corsAllowOrigin: "https://alexandria.stud.k8s.aet.cit.tum.de"
 
 # NetworkPolicies: default-deny ingress with explicit allow rules.
 # Off by default.
--- a/services/client/README.md
+++ b/services/client/README.md
@@ -18,6 +18,26 @@ rendered with Markdown support. Each answer includes references to the source
 documents it was derived from; clicking a source reference switches back to
 the **Documents** tab and opens that document.
 
+Configuration
+-------------
+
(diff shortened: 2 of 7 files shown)


infra/k8s/alexandria/values.yaml
```diff
@@ -148,6 +148,10 @@ ingress:
   # nginx client_max_body_size, kept a bit above Spring MAX_UPLOAD_SIZE
   # so oversized uploads get the Spring 413 error response
   maxBodySize: "30m"
+  # Browser origin allowed to call the API cross-origin. Same-origin traffic
+  # through this ingress is unaffected; override when the client is hosted
+  # elsewhere.
+  corsAllowOrigin: "http://localhost"
```
BjarneHa: Why does this default to localhost?

services/client/README.md
```diff
@@ -18,6 +18,26 @@ rendered with Markdown support. Each answer includes references to the source
...
+Configuration
+-------------
+
+The client reads a single build-time variable:
+
+| Variable       | Default | Purpose                                                                 |
+| -------------- | ------- | ----------------------------------------------------------------------- |
+| `VITE_API_URL` | `""`    | Base URL prefixed to every API request made by the generated SDK client |
+
+Vite inlines `VITE_*` variables at build time, so this is set when the image is
```
BjarneHa, line 29: Debatable how useful this is when it is inserted at build time. However, since the guidelines on Artemis specify that we should use build time `VITE_` variables, this is fine.

infra/k8s/alexandria/values.yaml
DoPri: See "Notes for reviewer" section in the PR description

services/client/README.md
DoPri: I agree. As you said, I added it only because the guidelines say so.

infra/k8s/alexandria/values.yaml
BjarneHa: It's not a no-op though. If this is deployed on e.g., the AET cluster, cross-origin access from a local webserver is still allowed if I'm not mistaken. There's no reason for this to be the case.

DoPri: Should be fixed now. Please verify. @BjarneHa

BjarneHa (approved):
Since the deployment is fixed to the AET cluster anyway, this is fine.

PR #339, opened 2026-07-16
Feat: track async pipeline status for summary, entities and tags

BjarneHa:
## What does this PR do?

After uploading a document the client had no way to tell whether the GenAI pipeline was still running or had failed -- both cases looked identical (null summary, empty entities/tags). Closes #264.

The fix tracks pipeline state explicitly. A `PENDING` row is written synchronously before the async job fires, so the first fetch after upload already carries a status. The async job then flips it to `COMPLETED` or `FAILED`.

Summary state lives on the `Summary` row itself (`status`, `errorMessage`). Entities and tags have no aggregate row, so their state lives as `entitiesStatus` / `tagsStatus` columns on `Document`.

On the client side, all three pending states render a loading indicator instead of hiding the section entirely. The documents list and the detail view both poll every 3s while any status is `PENDING` and stop automatically once everything settles.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] CI / infra

## Definition of Done

- [ ] CI is green

## Notes for the reviewer

Two Flyway migrations:
- `V2__summary_status.sql` -- adds `status` and `error_message` to `summaries`, relaxes `NOT NULL` on content/model/generated_at (null while pending/failed). Existing rows default to `COMPLETED`.
- `V3__document_processing_status.sql` -- adds `entities_status` and `tags_status` to `documents`. Existing rows default to `COMPLETED`.

`SummaryStatus` (PENDING/COMPLETED/FAILED) is reused for all three fields rather than introducing a second identical enum.

The `reprocessSummary` path reuses the existing `Summary` row and resets it to `PENDING` rather than deleting and recreating it, so there is no gap in visibility during a reprocess.

## Summary by CodeRabbit

* **New Features**
  * Added processing state and error details for document summaries, extracted entities, and tags (pending, completed, failed).
  * Added animated loading indicators in the document list and detail view while processing is in progress.
  * Document details now auto-refresh periodically until summary, entities, and tags complete.
  * When generation fails, the UI displays failure messaging and avoids showing incomplete results.
* **API / Data**
  * Updated API responses and OpenAPI schema to include new status fields and optional error messages for summaries and processing stages.


services/spring/knowledgebase-service/src/main/java/com/alexandria/knowledgebase/KnowledgeBaseService.java
```diff
@@ -211,33 +260,77 @@ public List<Document> resolveDocuments(String ownerSubject, Collection<String> o
...
+    }
+
+    @Async
+    @Transactional
+    public void reprocessTagsAsync(UUID id) {
+        Document document = findDocumentForAsyncStep(id);
+        if (document != null) {
+            processTags(document);
+        }
+    }
```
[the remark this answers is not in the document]
ladu-tu: This is in theory correct, but would only trigger on rapid double-reprocess. As this would need a lot of rework, we won't address this here.

services/spring/knowledgebase-service/src/main/java/com/alexandria/knowledgebase/KnowledgeBaseService.java
```diff
@@ -161,9 +205,12 @@ private void processEntities(Document document) {
...
                 document.getExtractedEntities().add(entity);
             }
+            document.setEntitiesStatus(SummaryStatus.COMPLETED);
+            documentService.save(document);
         } catch (Exception e) {
-            log.warn(
-                    "GenAI entity extraction failed for document {}: {}", document.getId(), e.getMessage());
+            log.warn("GenAI entity extraction failed for document {}: {}", document.getId(), e.getMessage());
+            document.setEntitiesStatus(SummaryStatus.FAILED);
+            documentService.save(document);
```
DoPri: `processEntities` and `processTags` each do a read-modify-write of the whole `Document` via `documentService.save(document)`. `Document` has no `@Version` and no `@DynamicUpdate` (`document/Document.java:13`), so Hibernate writes every column on update. If `reprocessEntitiesAsync` and `reprocessTagsAsync` run at the same time (two reprocess clicks, or a reprocess overlapping the initial pipeline) they both load the row, then the last commit overwrites the other's status field with its stale snapshot. One of `entities_status` / `tags_status` gets reverted to PENDING and never recovers, so the client polls that document forever. Update only the status column (derived update query) or add `@DynamicUpdate` / `@Version`.

services/spring/knowledgebase-service/src/main/java/com/alexandria/knowledgebase/KnowledgeBaseService.java
```diff
@@ -143,13 +166,34 @@ public void processDocumentAsync(UUID documentId) {
     }
```
DoPri, line 189: `processDocumentAsync` wraps summarize + entities + tags + index in one `@Transactional`. None of the intermediate COMPLETED/FAILED writes are visible until the whole method commits, so on the main upload flow the client sees all-PENDING and then a single flip at the end. The per-step statuses only actually work for the reprocess paths (separate transactions). Same transaction boundary means a pod restart or an uncaught error mid-pipeline rolls everything back to the committed PENDING state with no retry, leaving the document stuck. Split the steps into separate transactions (like the reprocess workers already do) so each status commits on its own.

services/client/src/components/MainView.tsx
```diff
@@ -16,7 +16,19 @@ export default function MainView() {
...
+      if (!docs) return false;
+      const anyPending = docs.some(
+        (doc) =>
+          doc.summary?.status === "PENDING" ||
+          doc.entitiesStatus === "PENDING" ||
+          doc.tagsStatus === "PENDING",
+      );
+      return anyPending ? 3000 : false;
+    },
+  });
```
DoPri, line 29: Polling only stops when nothing is PENDING. Every path that can strand PENDING (issues 1 and 2, or `findDocumentForAsyncStep` rethrowing on a transient DB error at `KnowledgeBaseService.java:328-338`) leaves the UI spinning and hitting the API every 3s indefinitely, with no failure ever shown to the user. The list poll also re-serializes all documents each tick, and each doc lazy-loads its summary (N+1) via open-in-view. Add a cap (max attempts or max age) that stops polling and surfaces a failed/timed-out state, or time out stuck PENDING rows to FAILED server-side

Note: Same for `services/client/src/components/DocumentDetail.tsx:35-43`

DoPri (changes requested):

DoPri:
And also fix the CI and the merge conflicts, please ;)

services/spring/knowledgebase-service/src/main/java/com/alexandria/knowledgebase/KnowledgeBaseService.java
ladu-tu: Good call. Fixed.

services/spring/knowledgebase-service/src/main/java/com/alexandria/knowledgebase/KnowledgeBaseService.java
ladu-tu: Fixed.

services/client/src/components/MainView.tsx
ladu-tu: I made the client stop polling after a fixed number of retries. Changing this on the Spring side would have required a larger rework with db schema migration etc.

DoPri (approved):
Thank you

PR #343, opened 2026-07-16
CI: Retry Keycloak admin API calls in E2E provisioning

DoPri:
## What does this PR do?

Makes the "Provision Test User via Keycloak Admin API" step in the E2E job resilient to the startup flakiness that keeps failing it.

`docker compose --wait` gates on Keycloak's healthcheck, but healthy only means the realm import finished. For a few seconds after that, Traefik routing and the admin REST API can still refuse connections or return transient 5xx, and the very first curl (the admin token fetch) then kills the whole E2E job.

Each curl in the step now runs with a shared retry set:

```
--retry 5 --retry-all-errors --retry-connrefused --retry-delay 2 --max-time 30
```

So connection refusals and transient 5xx/timeouts get retried instead of failing on the first hiccup. Same flag style we already use for the coverage gist publish step. The flags live in a bash array (`"${RETRY[@]}"`) so shellcheck stays happy.

E2E still fails the pipeline if the tests genuinely fail. This only hardens the provisioning around Keycloak coming up.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] CI / infra

## Definition of Done

- [ ] CI is green (actionlint passes)
- [ ] If the API changed: `api/openapi.yaml` is updated and lint passes (n/a)
- [ ] If a service was added or restructured: docker-compose still comes up cleanly (n/a)
- [ ] Tests added or updated for the changed behaviour (n/a, CI reliability change)
- [x] Docs updated where it matters (inline comment explains why the retries are there)

## Notes for the reviewer

Kept the existing logic intact (still exits on a non-201/409 create, still reads the user id, still sets the password). Only the curl invocations changed. If provisioning still flakes after this we can escalate to a bounded polling loop on the token endpoint, but retries should cover the observed failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved end-to-end test user provisioning by automatically retrying transient connection failures.
  * Increased reliability when obtaining authentication tokens, creating test users, locating user records, and resetting passwords.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -698,8 +698,10 @@ jobs:
 
       - name: Provision Test User via Keycloak Admin API
         run: |
+          RETRY=(--retry 5 --retry-all-errors --retry-connrefused --retry-delay 2 --max-time 30)
+
           # Obtain an admin access token from the master realm
-          ADMIN_TOKEN=$(curl -fsS \
+          ADMIN_TOKEN=$(curl -fsS "${RETRY[@]}" \
             -d "client_id=admin-cli" \
             -d "username=${KC_BOOTSTRAP_ADMIN_USERNAME}" \
             -d "password=${KC_BOOTSTRAP_ADMIN_PASSWORD}" \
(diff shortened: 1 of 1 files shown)


ladu-tu (approved):
LGTM.

PR #355, opened 2026-07-18
Chore: GenAI hygiene and Spring->GenAI read timeout

DoPri:
## What does this PR do?

Cleanup around the GenAI service plus one cross-service timeout fix. No behaviour change to the AI endpoints themselves.

- **Dockerfile cleanup.** Dropped the three declared-but-unused build args (`RELEASE_TYPE`, `APP_VERSION`, `IMAGE_TAG`), the commented-out `# ENTRYPOINT ["sleep", "600"]` debug line, and the duplicate `WORKDIR /app`. Also pinned both base images by digest (`ghcr.io/astral-sh/uv:python3.14-alpine` and `python:3.14-alpine`), which is what the client and Spring Dockerfiles already do.
- **pyproject dedupe + version.** `dependencies` listed `uvicorn`, `pydantic`, and `langchain-openai` twice; removed the duplicates. Bumped the version from 5.0.0 to 6.0.0 so genai lines up with the Spring services (also updated `app/main.py`'s `SERVICE_VERSION`, which feeds the FastAPI version and the `application_version` metric, and the `alexandria-genai` entry in `uv.lock`).
- **Spring to GenAI read timeout.** Both `GenAiClient`s (knowledgebase- and qa-service) used a 30s read timeout, but the GenAI chat budget is `LLM_TIMEOUT_SECONDS`=60 with 2 retries, i.e. up to 180s worst case. So Spring could abandon a long-document request at 30s while GenAI kept working (and spending tokens) on something nobody was waiting for. Raised the read timeout to 200s so it sits above GenAI's worst case, rather than shortening GenAI, so the long-document requests the system is built for don't get cut off mid-flight.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Docs
- [ ] CI / infra

## Definition of Done

- [ ] CI is green

## Notes for the reviewer

- On the timeout: 200s covers the chat model's worst case (60s x 3). The qa `/ask` path does an embedding call plus the chat call, so in a pathological double-worst-case (both maxing out their retries) it could exceed 200s, but that only happens if GenAI is already badly degraded. If you'd rather cap the total instead, the alternative was shortening GenAI's own budget; I went with raising Spring per our discussion.

### DIFF
--- a/services/genai/Dockerfile
+++ b/services/genai/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/astral-sh/uv:python3.14-alpine AS builder
+FROM ghcr.io/astral-sh/uv:python3.14-alpine@sha256:7a6d8c4ff34812917124c77a8559458488f09d5e52e7646c04599f007956966c AS builder
 
 ENV PYTHONFAULTHANDLER=1 \
     PYTHONUNBUFFERED=1 \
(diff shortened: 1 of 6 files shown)


ladu-tu (commented):
Good catch regarding the Spring timeout!

PR #356, opened 2026-07-18
CI: Pin Azure deploy to commit SHA, prune images, make Prometheus private

DoPri:
## What does this PR do?

Brings the Azure VM deploy up to the same rigor the Kubernetes path already has, and closes an open Prometheus on the public deployment.

**Pin deployed images to the commit SHA + prune old ones.** The k8s deploy passes `image.tag=<sha>` to Helm, but the VM just did `docker compose pull` against `ghcr.io/...:latest`, so it ran "whatever latest was at pull time": not traceable to a commit, racy when two merges queue, and impossible to roll back precisely.

- `docker-compose.yml` now uses `:${IMAGE_TAG:-latest}` for the five app images, so it still defaults to `:latest` for local/manual use.
- The Ansible deploy role passes `IMAGE_TAG` (from the new `image_tag` var) into `docker compose pull` and `up`, defaulting to `latest` when unset.
- The CI `Deploy to Azure VM` job passes the built short SHA (`needs.build-image.outputs.image-tag`, the same value k8s already uses) as `image_tag`.
- The role also prunes unused images older than a week (`docker image prune -af --filter until=168h`) after each deploy, so old SHA-tagged layers stop piling up until the disk fills.

**Keep Prometheus off the public router.** On the public (Let's Encrypt) deployment, `/prometheus` was reachable unauthenticated: the TLS overlay only added TLS, no auth. In k8s Prometheus is private (port-forward only) and only Grafana is exposed (behind Keycloak). The overlay now disables the Prometheus Traefik router and publishes it on the VM's loopback (`127.0.0.1:9090`), so it's reachable via an SSH tunnel (`ssh -L 9090:localhost:9090 <vm>`), mirroring `kubectl port-forward`. Local dev is unchanged: it doesn't use the overlay, so `http://localhost/prometheus/` still works.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [x] CI / infra

## Definition of Done

- [ ] CI is green

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -24,6 +24,17 @@
 # PUBLIC_BASE_URL="http://localhost"
 
 # ----------------------------------------------------------------------------------
+# Image tag (deployment)
+# ----------------------------------------------------------------------------------
+
+# Which image tag docker-compose pulls for the app services (client, the three
+# Spring services, genai). Defaults to "latest". The Azure VM deploy pins this to
+# the commit SHA in CI so a deploy is traceable to a commit and can be rolled back
+# precisely; local dev leaves it at latest.
+
+# IMAGE_TAG="latest"
+
+# ----------------------------------------------------------------------------------
 # TLS / Let's Encrypt (docker-compose.letsencrypt.yml overlay)
 # ----------------------------------------------------------------------------------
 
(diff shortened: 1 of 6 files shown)


ladu-tu (approved):
Thanks for all the cleanup work!

